PR #5, opened 2026-05-29
Fix Application.cfc lifecycle target page

Blute:
## Summary

This adds Lucee-compatible target-page handling for `Application.cfc` request lifecycle methods.

When a request is executed from a physical source path under the web root, `onRequestStart(targetPage)`, `onRequest(targetPage)`, and `onRequestEnd(targetPage)` should receive the web-root-relative page path, e.g. `/_moopa.cfm`, rather than the VM/source path, e.g. `/app/_moopa.cfm`.

This came up while porting Moopa Starter, which uses `onRequest(targetPage)` as part of its front-controller dispatch flow.

## Changes

- Adds `CfmlVirtualMachine::lifecycle_target_page()` to normalize lifecycle target-page arguments against `ServerState.webroot`.
- Uses that normalized value for `onRequestStart`, `onRequest`, and `onRequestEnd`.
- Adds regression coverage for all three lifecycle methods receiving the same web-root-relative target page.

## Tests

- `cargo test -p cfml-vm --test lifecycle_target_page`
- `cargo test -p cfml-vm --test lazy_session`

### DIFF
--- a/crates/cfml-vm/src/lib.rs
+++ b/crates/cfml-vm/src/lib.rs
@@ -11011,6 +11011,28 @@ impl CfmlVirtualMachine {
         }
     }
 
+    /// Return the page path exposed to Application.cfc lifecycle methods.
+    ///
+    /// The VM keeps `source_file` as the physical path for include and
+    /// component resolution, but CFML engines pass web-root-relative paths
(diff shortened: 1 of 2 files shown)


Blute:
Closing this draft in favour of #6, which uses the project-focused branch name lucee-compat/application-lifecycle-target-page.

PR #9, opened 2026-05-31
Preserve nested argument references

Blute:
## Summary
- Adds a CFML runner regression for nested argument references.
- Preserves writeback when an argument aliases a nested caller path.
- Covers the same behavior with a Rust VM regression test.

## CFML repro
- tests/core/test_argument_reference_nested.cfm

## Tests
- cargo test -p cfml-vm --test argument_reference

### DIFF
--- a/crates/cfml-codegen/src/compiler.rs
+++ b/crates/cfml-codegen/src/compiler.rs
@@ -190,6 +190,7 @@ pub enum BytecodeOp {
     /// TRUE arm (back to body); falling through means the loop has finished.
     ForLoopStep(String, i64, CmpOp, i64, usize),
     Call(usize),
+    CallWithArgSources(usize, Vec<Option<Vec<String>>>),
     Return,
 
     // Collections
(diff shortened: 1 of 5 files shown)


alexskinner:
Thanks for this — really thorough, especially the test coverage.
After review, the behaviour this PR targets (mutating a nested struct passed as an argument writing back to the caller’s path) is already correct on main. It landed via the reference-type work that came after this branch was cut:
	∙	PR-1 — arrays as reference types (dd85c19)
	∙	PR-2 — structs as reference types (701b889)
Structs are now backed by a shared Arc<RwLock<IndexMap>>, so a passed-in nested struct — and any aliases of it — observe mutations natively, without explicit writeback.
I verified all five scenarios from your test_argument_reference_nested.cfm pass unmodified on current main, and confirmed the same against Lucee 7 — 5/5 on both engines. So the VM-level writeback / alias-replacement machinery in this PR is no longer needed. It would also conflict with the current type model, since the branch predates the CfmlValue boxing rework (PR-A).
The regression test itself is genuinely valuable, so I’ve adopted it (credited to this PR) on branch adopt-nested-argref-test. Closing this in favour of that — thanks again!

PR #11, opened 2026-05-31
Support cfloop collection item index

Blute:
## Summary
- Adds a CFML runner regression for collection loops using both item and index.
- Makes collection loops expose the item value and key/index consistently with Lucee behavior.

## CFML repro
- tests/core/test_cfloop_collection_item_index.cfm

## Tests
- cargo test -p cfml-vm --test cfloop_collection

### DIFF
--- a/crates/cfml-compiler/src/tag_parser.rs
+++ b/crates/cfml-compiler/src/tag_parser.rs
@@ -397,7 +397,7 @@ fn parse_cf_tag(chars: &[char], start: usize, len: usize, imports: &mut std::col
             ("} else {\n".to_string(), tag_end - start)
         }
         "cfloop" => {
-            parse_cfloop_tag(&attrs, tag_end - start)
+            parse_cfloop_tag(&attrs, chars, tag_end, len, imports, tag_end - start)
         }
         "cfscript" => {
             // Everything between <cfscript> and </cfscript> is raw script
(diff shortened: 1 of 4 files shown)


alexskinner:
Thanks for spotting this — the gap is real: <cfloop collection item index> does throw Variable is undefined on main, because index wasn’t accepted as a key alias for collection loops.
On review I went with a smaller fix. This branch predates the reference-type work that landed afterward (arrays PR-1 dd85c19, structs PR-2 701b889), and that changes what’s needed:
	∙	The manual writeback (collection[key] = item each iteration) is redundant. Structs are now reference types backed by a shared Arc<RwLock<IndexMap>>, so mutating item.field inside the loop already propagates to the collection — no write-back needed. Verified on current main.
	∙	It also diverges from Lucee for scalar items. I probed Lucee 7 across the attribute matrix: reassigning a scalar item (<cfset v = "changed">) must not write back into the collection — Lucee keeps the original value. The unconditional collection[key] = item would write "changed" back, which Lucee never does. The PR’s tests only cover struct mutation, so this case slips through.
So the actual gap is just the missing index→key alias, which is a one-liner:

let key = attrs.get("key").or_else(|| attrs.get("index"));

I’ve landed that on main (db6a6a2), and adapted your struct regression cases (credited to this PR) plus added item-only and scalar-reassign guards so the full Lucee matrix is pinned down — 4/4 on both RustCFML and Lucee 7. Closing this in favour of that fix. Thanks again for the catch and the test coverage!

PR #12, opened 2026-05-31
Support negative cfloop steps

Blute:
## Summary
- Adds a CFML runner regression for counted cfloop with a negative step.
- Updates counted loop translation so descending ranges execute correctly.

## CFML repro
- tests/core/test_cfloop_negative_step.cfm

## Tests
- cargo test -p cfml-vm --test cfloop_negative_step

### DIFF
--- a/crates/cfml-compiler/src/tag_parser.rs
+++ b/crates/cfml-compiler/src/tag_parser.rs
@@ -1824,10 +1824,15 @@ fn parse_cfloop_tag(
         let from = strip_hashes(from);
         let to = strip_hashes(to);
         let step = strip_hashes(&step);
+        let comparison = if step.trim_start().starts_with('-') {
+            ">="
+        } else {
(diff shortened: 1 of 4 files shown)


alexskinner:
Thanks for this! Verified the bug and your fix — the counted-loop translation does hardcode <=, so descending ranges never run.
One gap: the step.starts_with('-') check is textual, so it only catches literal negative steps. A runtime-valued step slips through and still diverges from Lucee:

I’ve landed an alternative on main (commit) that decides direction at runtime instead — for (var i = from; (step < 0 ? i >= to : i <= to); i = i + step) — which handles both literal and dynamic steps and matches Lucee on all cases. Kept your test and added dynamic-step / equal-bounds / magnitude coverage.
Going to close this in favour of that, but the report and repro were spot on — much appreciated 🙏

PR #15, opened 2026-05-31
Resolve leading slash component mappings

Blute:
## Summary
- Adds a CFML runner regression for createObject() with a leading-slash component mapping path.
- Resolves mapped component paths such as /oop/ComponentName consistently with Application.cfc mappings.

## CFML repro
- tests/oop/test_component_mapping_paths.cfm

## Tests
- cargo test -p cfml-vm --test component_mapping

### DIFF
--- a/crates/cfml-vm/src/lib.rs
+++ b/crates/cfml-vm/src/lib.rs
@@ -9132,8 +9132,17 @@ impl CfmlVirtualMachine {
         if self.mappings.is_empty() {
             return None;
         }
-        // Convert dot-path to slash-path: "taffy.core.api" → "/taffy/core/api"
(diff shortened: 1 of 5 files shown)


alexskinner:
My fix (broader than the PR's): The PR routed leading-slash paths through mappings only — but its
  own CFML test defines no mappings, so that path alone wouldn't resolve it. The correct CFML
  semantics: a leading-slash component path is webroot/mapping-relative, exactly like a
  leading-slash <cfinclude>. So I route it through the existing resolve_leading_slash_include chain
  (configured mappings → serve-mode webroot → entry-template directory), only falling back to
  literal-path treatment if a genuinely OS-absolute .cfc actually exists. This subsumes the PR's
  mapping case and handles the no-mapping case the CFML test needs.

PR #16, opened 2026-05-31
Bind component method named arguments

Blute:
## Summary
- Adds a CFML runner regression for component method named arguments and argumentCollection.
- Binds component method calls by parameter name instead of call-site ordering.
- Covers both direct named arguments and argumentCollection with a Rust VM regression test.

## CFML repro
- tests/oop/test_component_method_named_args.cfm

## Tests
- cargo test -p cfml-vm --test component_method_named_args


alexskinner:
Did a broader review and push

PR #17, opened 2026-05-31
Prefer component methods over struct helpers

Blute:
## Summary
- Adds a CFML runner regression for component methods named like struct member helpers.
- Dispatches to the component method before falling back to struct helper behavior.
- Covers delete() and count() method names with a Rust VM regression test.

## CFML repro
- tests/oop/test_component_method_precedence.cfm

## Tests
- cargo test -p cfml-vm --test component_method_precedence


alexskinner:
have taken the test but struct helpers should never be used on components instead it should trigger on missingmethod and then error if none found

PR #18, opened 2026-05-31
Implement applicationStop lifecycle reset

Blute:
## Summary
- Adds a lifecycle CFML fixture for applicationStop() across requests.
- Implements applicationStop() so it clears shared application scope, marks the app unstarted, and drops cached lifecycle functions.
- Prevents the end-of-request application writeback from restoring the stopped state.

## CFML repro
- tests/lifecycle/application_stop/Application.cfc
- tests/lifecycle/application_stop/index.cfm
- tests/lifecycle/application_stop/stop.cfm

## Tests
- cargo test -p cfml-vm --test application_stop

### DIFF
--- a/crates/cfml-vm/src/lib.rs
+++ b/crates/cfml-vm/src/lib.rs
@@ -520,6 +520,10 @@ pub struct CfmlVirtualMachine {
     pub request_scope: IndexMap<String, CfmlValue>,
     /// Application scope — shared across requests (Arc<Mutex> for thread safety)
     pub application_scope: Option<Arc<Mutex<IndexMap<String, CfmlValue>>>>,
+    /// Name of the application currently attached to application_scope.
+    current_application_name: Option<String>,
+    /// Set when applicationStop() has cleared the attached application during this request.
+    application_stopped: bool,
     /// Server-level state — persists across requests in --serve mode
     pub server_state: Option<ServerState>,
     /// HTTP response headers set by cfheader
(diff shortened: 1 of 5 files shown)


alexskinner:
PR #18 (applicationStop) — landed as 1807f87, pushed to main

  The bug: applicationStop() was a no-op stub (builtins.rs → Ok(Null)). The application never stopped. Reproduced via the PR's
  Rust integration test, which failed at "must mark the application unstarted."

  The fix (in crates/cfml-vm/src/lib.rs): made applicationStop VM-intercepted. stop_current_application() now:
  - clears the shared ApplicationState (variables, started=false, all three function caches),
  - clears the live application scope,
  - sets a flag so the end-of-request writeback can't resurrect the cleared state.

  What the stale-base PR missed — and I fixed:
  1. cached_function_indices — a field added to ApplicationState after the PR branched. The PR didn't clear it, leaving stale
  indices. Mine does (test asserts it's empty).
  2. onApplicationEnd firing — my Lucee 6.1.4 probe showed Lucee invokes onApplicationEnd(applicationScope) synchronously
  during applicationStop(), before teardown (lifecycle.log = START/END/START). The PR omits this. After you approved, I added
  it via the existing onSessionEnd pattern. RustCFML serve mode now reproduces Lucee's START/END/START exactly.

  Left alone (you approved): unset application.x returns empty in RustCFML vs Lucee's key doesn't exist throw — confirmed this
  is general scope-access behavior (any unset key, any scope), not stop-specific. Flagged for a separate task.

  Test: kept the PR's Rust regression test and extended it — asserts the restart generates a fresh seed (proving
  onApplicationStart genuinely re-ran), that onApplicationEnd fires exactly once with the pre-stop scope, and that it does not
  fire on ordinary requests or on restart.

PR #22, opened 2026-05-31
Add cross-engine test for the `component` reserved-keyword parse gap

bpamiri:
## Summary

A cross-engine test for the CFML behavior that is currently the **terminal blocker** to serving a Wheels request on RustCFML: the `component` token is treated as a hard reserved keyword, so `cfinvoke component=…` (and a bare `component = x`) fail to parse. **Passes on Lucee 7, fails on RustCFML 0.34.3.**

Contained in a runtime-instantiated fixture so the parse failure degrades only that component (to a non-object) instead of aborting the run.

## Context — Wheels boots, then this is the last wall 🎉

With #21's auto-vivify fix, Wheels now parses `Application.cfc`, runs its full lifecycle, and bootstraps DI on RustCFML. This is the next (and, as far as I can tell, last big) gap before it can actually serve a page.

## The gap

`component` is a **soft** keyword on Lucee 5/6/7, Adobe CF 2018–2025, and BoxLang — it introduces a CFC only when it *begins a declaration*, and is otherwise a normal identifier (legal as a variable name and as `cfinvoke`'s `component` attribute). RustCFML treats it as a **hard** reserved keyword. Any place the token `component` is followed by `=` fails to parse:

```
Parse error: Expected LBrace, found Equal
```

```cfml
component = "x";              // bare assignment → parse error

cfinvoke                      // script-statement attribute → parse error
component = "Svc"
method = "m";

cfinvoke(component="Svc", method="m")   // function-CALL form → parses fine ✅
```

The function-call form parsing fine (while the statement form fails) should help scope the fix, and is a viable Wheels-side workaround.

## Why it's the terminal blocker

`public/…/Global.cfc::$cfinvoke()` is written in the exact failing shape (`cfinvoke` / `component = "…"` / …). So:

1. `wheels.Global` fails to parse.
2. RustCFML degrades an unparseable component to a **non-object, silently** (no thrown error).
3. `createObject("component","wheels.Global")` → non-object → the DI Injector's `getInstance("global")` hands that to `application.wo`.
4. Every lifecycle call on `application.wo` (`$cgiScope`, `$initializeRequestScope`, `$runOnRequestStart`) **silently no-ops**.
5. `request.cgi` / `request.wheels` never get set → the dispatcher resolves an empty path → **empty 200, no error**.

I chased this through the request scope and the DI singleton before instrumenting `$initializeRequestScope` from the inside and finding it simply never runs — `application.wo` is a non-object because `wheels.Global` won't parse, and it all traces to this one token.

## Validation

| Engine | Result |
|---|---|
| Lucee 7.0.2 | **1/1 pass** (fixture parses, cfinvoke runs) |
| RustCFML 0.34.3 | **0/1** (fixture degrades to a non-object) |

### DIFF
--- a/tests/runner.cfm
+++ b/tests/runner.cfm
@@ -209,6 +209,12 @@ try { include "core/test_forin_member_loop_var.cfm"; } catch (any e) { writeOutp
 //     for everything except the `local` scope. Wrapped in try/catch so the
 //     throw fails its assertions without aborting the run.
 try { include "core/test_undefined_var_autovivify.cfm"; } catch (any e) { writeOutput("ERROR | core/test_undefined_var_autovivify.cfm | " & e.message & chr(10)); }
+//   - cfinvoke_component_attr: `component` is a soft keyword on Lucee/ACF/BoxLang
+//     but a hard reserved keyword on RustCFML, so `cfinvoke component=...` (and a
+//     bare `component = x`) fail to PARSE. Wheels' Global.cfc::$cfinvoke() uses
+//     this exact shape, so wheels.Global degrades to a non-object and the whole
+//     framework boots-but-serves-empty. Parse failure contained in a fixture.
+try { include "tags/test_cfinvoke_component_attr.cfm"; } catch (any e) { writeOutput("ERROR | tags/test_cfinvoke_component_attr.cfm | " & e.message & chr(10)); }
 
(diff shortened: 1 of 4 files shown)


alexskinner:
ive taken the elements and tried to incorporate

PR #27, opened 2026-05-31
Add CFML test for preserveSingleQuotes

Blute:
## Summary
- Adds a CFML runner test for Lucee-compatible `preserveSingleQuotes()` behaviour.
- Moopa uses this helper around SQL fragments that contain quoted literals and PostgreSQL casts.

## Expected Behaviour
`preserveSingleQuotes(value)` should return the original string unchanged while marking embedded single quotes as preserved for query handling, matching Lucee compatibility expectations.

## Current RustCFML Behaviour
Running the runner on current `main` reports the missing builtin:

```text
ERROR | stdlib/test_preserve_single_quotes.cfm | Variable 'preserveSingleQuotes' is undefined
```

## Test


alexskinner:
Manually fixed

PR #28, opened 2026-05-31
Add CFML test for keyword member loop targets

Blute:
## Summary
- Adds a CFML runner test for Lucee-compatible dotted loop targets whose final member name is a keyword.
- Moopa route metadata can use names such as `local.package`; Lucee allows this as ordinary member access.

## Expected Behaviour
A script `for ... in` loop should allow assigning each item into a dotted variable target like `local.package`, even though `package` is a keyword-like identifier.

## Current RustCFML Behaviour
Running the runner on current `main` reaches the new final repro and fails while parsing the loop target:

```text
Parse error in tests/core/test_forin_keyword_member_access.cfm [line 6, col 24]: Expected Semicolon, found In
```

## Test

### DIFF
--- a/tests/core/test_forin_keyword_member_access.cfm
+++ b/tests/core/test_forin_keyword_member_access.cfm
@@ -1,0 +1,13 @@
+<cfscript>
+suiteBegin("For-in keyword member access");
+
+local = {};
+items = ["ok"];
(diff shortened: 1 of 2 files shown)


alexskinner:
manual merge of test and fixed

PR #30, opened 2026-05-31
Add CFML test for soft-keyword function names

Blute:
## Summary
- Adds a CFML runner test for Lucee-compatible component function names that use soft-keyword identifiers.
- Moopa can define route-style component methods named `new`, and Lucee preserves/calls those methods normally.

## Expected Behaviour
A `<cffunction name="new">` method should appear in `getMetadata(component).functions` as `new` and should be callable as `component.new()`.

## Current RustCFML Behaviour
Running the runner on current `main` reaches the new test but the method is neither present in metadata nor callable:

```text
FAIL | Soft-keyword function names | 0/2 passed (2 failed)
```

## Test


alexskinner:
close manually and fixed

PR #32, opened 2026-05-31
Add cross-engine tests for reserved-word identifiers and quoted catch types

bpamiri:
Continuing the Wheels-on-RustCFML compatibility work (follow-on to #24). While booting the Wheels framework on **v0.37.0**, three more constructs that Lucee 5/6/7, Adobe CF 2018-2025, and BoxLang all accept still fail to parse on RustCFML — and because the enclosing component then degrades to a non-object silently, they're invisible until you ask `isObject()`.

This PR adds **two cross-engine suites** (+ 3 fixtures).

## Suite 1 — reserved words usable as identifiers

A direct follow-on to the v0.36.0 *soft `component` keyword* work: two more words that are soft keywords on the JVM engines but hard-reserved on RustCFML.

**`new` as a function name.** `new Foo()` is the instantiation operator, but `new` is equally legal as a method name — and Wheels' core creation API depends on it:

```cfml
component { public any function new() { /* ... */ } }   // FAIL: Expected identifier, found New
```

`model("User").new()` is backed by `public any function new(...)` in `vendor/wheels/model/create.cfc`.

**`extends` / `implements` as parameter names:**

```cfml
function generateClass( string extends="", string implements="" ) {}   // FAIL: Expected RParen, found Extends
```

Used verbatim in `vendor/wheels/wheelstest/system/mockutils/MockGenerator.cfc`.

## Suite 2 — quoted string catch type

A catch clause may name the exception type as a quoted string (the idiomatic way to catch a dotted, namespaced custom exception):

```cfml
try { throw(type="My.Custom.Type", message="x"); }
catch ("My.Custom.Type" e) { /* ... */ }                // FAIL: Expected identifier, found String(...)
```

**This is the next blocker on the Wheels boot path after #24's header gaps.** `vendor/wheels/Public.cfc:56` (`catch ("Wheels.Packages.RegistryUnavailable" e)`) is instantiated unconditionally at `onApplicationStart` in development mode; `vendor/wheels/auth/JwtStrategy.cfc` does the same with `"Wheels.Auth.JWT.TokenExpired"`.

## Verification

| Engine | Result |
|--------|--------|
| Lucee 7.0.2 | **3/3 pass** |
| RustCFML 0.37.0 | 0/3 — all three degrade to a non-object |

Failing declarations live in runtime-instantiated **fixtures** (a parse error escapes `try/catch` and would abort the runner); via `createObject` they degrade to a non-object so the assertions surface the gap and the full runner still completes (3104/3107, the 3 intended new reds the only failures).

---

## Other parse gaps found while booting Wheels on 0.37 (not in this PR — signal only)

In addition to #24's roadmap (component `extends`-after-attribute & unquoted-boolean attrs — both still open here) and its noted items (interface `extends=`, dotted-FQN param types, `+=` in a `for` increment, `function f() output=true {...}`), the 0.37 sweep surfaced a few more — all fail on 0.37, pass on Lucee/ACF/BoxLang, happy to send tests for any:

- **Compound assignment in more positions** — `+=` inside a braced `switch`/`case` block (`case "x": { total += 1; }`), and a chained `&=` (`local.sql = local.sql &= ";"`, `vendor/wheels/migrator/Base.cfc`). Bare `+=`/`&=` statements parse fine; only these positions fail.
- **Script tag-with-body** — `transaction action="begin" { ... }` (cfscript transaction with attributes + body) → `Expected RBrace, found LBrace`; and `for (item in collection)` nested inside a `cfhttp(){ ... }` script-tag body → `Expected RParen, found In`.

(FYI, unchanged & not Wheels blockers: the script-statement `cfinvoke component=... method=...;` form added in 0.36 still works great; only the function-call form `cfinvoke(component="x", ...)` ignores the `component` named arg at runtime. `this.mappings`-based `cfinclude` remains fixed.)

### DIFF
--- a/tests/core/NewMethodFixture.cfc
+++ b/tests/core/NewMethodFixture.cfc
@@ -1,0 +1,17 @@
+// Fixture: a component with a method literally named `new`. On Lucee/Adobe CF/
+// BoxLang `new` is a SOFT keyword — it introduces the `new Foo()` operator, but
+// is equally legal as a function name, and calling it via `this.new()` works.
+// On RustCFML 0.37.0 `new` is a HARD reserved keyword, so `function new(){...}`
+// fails to parse ("Expected identifier, found New") and the component degrades
+// to a non-object.
+//
+// Wheels' core object-creation API depends on this: `model("User").new()` is
+// backed by `public any function new(...)` in vendor/wheels/model/create.cfc.
+component {
+	public string function new() {
+		return "made";
+	}
+	public string function probe() {
+		return this.new();
+	}
+}
--- a/tests/core/QuotedCatchFixture.cfc
+++ b/tests/core/QuotedCatchFixture.cfc
@@ -1,0 +1,21 @@
+// Fixture: a try/catch whose catch clause names the exception type as a QUOTED
+// string literal (a dotted custom type that is not a bare identifier). On
+// Lucee/Adobe CF/BoxLang the catch type may be written as a quoted string, which
+// is how you catch a namespaced custom exception. On RustCFML 0.37.0 the catch
+// parser expects a bare identifier and rejects the string literal ("Expected
+// identifier, found String(...)"), so the component degrades to a non-object.
+//
+// Wheels uses this on the boot path — e.g. vendor/wheels/Public.cfc:
+//   catch ("Wheels.Packages.RegistryUnavailable" e) { ... }
+// and vendor/wheels/auth/JwtStrategy.cfc:
+//   catch ("Wheels.Auth.JWT.TokenExpired" e) { ... }
+component {
+	public string function probe() {
+		try {
+			throw(type = "A.B.C", message = "boom");
+		} catch ("A.B.C" e) {
+			return "caught";
+		}
+		return "not-caught";
+	}
+}
--- a/tests/core/ReservedParamFixture.cfc
+++ b/tests/core/ReservedParamFixture.cfc
@@ -1,0 +1,16 @@
+// Fixture: a function whose parameters are named with the soft keywords
+// `extends` and `implements`. On Lucee/Adobe CF/BoxLang these are legal
+// parameter names (and reachable via the arguments scope). On RustCFML 0.37.0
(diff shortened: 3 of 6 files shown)


alexskinner:
```markdown
  Thanks @bpamiri — reviewed against current `main` (v0.40.0, well past the v0.37.0
  this was filed on), and the substance here is now resolved on main.

  **The three constructs this PR targets:**

  | Construct | Status on main |
  |---|---|
  | `new` as a method name | Already worked (landed in PR #30, `e61eaa9`); also covered by `tests/oop/test_soft_keyword_function_name.cfm` |
  | `extends`/`implements` as parameter names | **Fixed** in `d729219` (now soft keywords — usable as param names *and* bare identifiers) |
  | quoted-string catch type `catch ("My.Type" e)` | **Fixed** in `d729219` |

  **The "signal only" gaps you listed at the bottom** — re-tested on main and addressed:

  - chained `&=` (`local.sql = local.sql &= ";"`) → **fixed** (`d729219`)
  - `+=` inside a braced `case`/`default` body → **fixed** (`d729219`)
  - cfscript `transaction action="begin" { ... }` (attribute form) → **fixed** (`d729219`)
  - `+=` in a `for`-increment and `function f() output=true {}` → were already fixed before this
  - `for (x in coll)` inside a `cfhttp(){}` body → **not fixed here.** It needs runtime
    param-collection (the same rework `cfquery` got in `28af97d`) and affects the tag
    form too, so I've split it out as a separate known issue: <link>

  Also note the PR description's "#24 roadmap still open here" list (component
  `extends`-after-attribute, unquoted-boolean attrs, interface `extends=`, dotted-FQN
  param types) had all already landed before this PR — they're green on main.

  Your fixtures and both suites were adapted into the test tree (credited), alongside
  new suites for the chained-assign, braced-case, and transaction fixes. Everything is
  green on **RustCFML (3242/3242 across 378 suites)** and **Lucee 7.0.3**.

  Since the fixes are on main, I'll close this as superseded — but thank you, the
  fixtures and the precise repro write-ups made this fast to land.

PR #38, opened 2026-05-31
Add CFML test for cfdirectory in component method

Blute:
## Summary

Adds a CFML runner test for using `<cfdirectory>` inside a CFC method with an unscoped `name` target.

The fixture component calls:

```cfml
<cfdirectory action="list" directory="/oop/native_cfcs" name="qRoutes" filter="*.cfc">
<cfreturn qRoutes.recordCount & "|" & qRoutes.columnList>
```

## Why this matters

Moopa route/table loaders use `cfdirectory` from inside component methods and expect the named result to be available in the method's local/function scope as a normal Lucee query object.

## Current RustCFML result

On current upstream, the new assertion fails with:

```text
FAIL: cfdirectory unscoped name inside component method | expected: [3|name,directory,size,type,dateLastModified,attributes,mode] | got: [3|]
```

## Scope

### DIFF
--- a/tests/oop/CfdirectoryLoader.cfc
+++ b/tests/oop/CfdirectoryLoader.cfc
@@ -1,0 +1,8 @@
+<cfcomponent>
+
+    <cffunction name="loadRoutes" returntype="string">
+        <cfdirectory action="list" directory="/oop/native_cfcs" name="qRoutes" filter="*.cfc">
+        <cfreturn qRoutes.recordCount & "|" & qRoutes.columnList>
+    </cffunction>
+
+</cfcomponent>
--- a/tests/tags/test_cfdirectory_mapping_path.cfm
+++ b/tests/tags/test_cfdirectory_mapping_path.cfm
@@ -7,5 +7,8 @@ suiteBegin("cfdirectory mapping paths");
     filter="*.cfc">
 <cfscript>
 assert("mapped directory resolves via Application.cfc mapping", mappedCfcFiles.recordCount, 3);
+
+loader = createObject("component", "oop.CfdirectoryLoader");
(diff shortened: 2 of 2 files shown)


alexskinner:
Thanks @Blute! Landed an enhanced fix on main (a4ff01e) — but the investigation took a turn worth sharing.

  Bringing the test onto current main (your branch was based on a stale base, ~e1d1c45) the failure was different from the
  empty-columnList one you saw — on current main columnList is fully populated. Digging in, two things surfaced:

  1. There's no cfdirectory ordering bug. I probed Lucee@7 directly, and its cfdirectory columnList is
  NAME,SIZE,TYPE,DATELASTMODIFIED,ATTRIBUTES,MODE,DIRECTORY — i.e. directory last. RustCFML's order already matched that.
  The expected value in the test (name,directory,size,..., directory second + lowercase) didn't match Lucee or RustCFML — it
  was an artifact of the stale base where columnList returned empty.
  2. The genuine divergence was global, not cfdirectory-specific: query.columnList reported column names in their original
  case (e.g. queryNew("id,fullName").columnList → id,fullName), whereas both Lucee and Adobe CF uppercase them
  (ID,FULLNAME). RustCFML was the outlier here.

  So the fix uppercases columnList output at all five sites (the three query.columnList dot-access paths, the
  queryColumnList() builtin, and the worker's DB-query result), corrects the test's expected value to Lucee's actual output,
  and adds uppercase-columnList regression guards to the Query suite.

  Verified green on both engines: cfdirectory 13/13, Query 16/16, QueryColumn 15/15 identical on Lucee@7; full CLI suite
  3218/3218; wasm target builds clean.

  Closing in favour of the committed fix — really appreciate the cross-engine test, it flagged a real ACF/Lucee parity gap
  well beyond cfdirectory. 🙏

PR #45, opened 2026-05-31
Add CFML test for cfhttp attribute interpolation

Blute:
## Summary

Adds a CFML server-runner test for hash interpolation in quoted `<cfhttp>` attributes.

The test is skipped in plain CLI mode, like the existing HTTP statement subtests, and runs when `tests/runner.cfm` is served by RustCFML. It calls a local target page using:

```cfml
<cfhttp url="#baseUrl##dynamicPath#" method="#httpMethod#" result="dynamicHttpResult">
```

## Why this matters

Moopa and Lucee applications often build request URLs dynamically in tag attributes. Lucee treats quoted attributes with embedded `#...#` as interpolated strings; RustCFML should not collapse the expressions into one bare variable name.

## Current RustCFML result

When served via RustCFML on current upstream, the new assertions fail with:

```text
FAIL: tag cfhttp interpolated attributes error | expected: [] | got: [Variable 'baseUrldynamicPath' is undefined]
FAIL: tag cfhttp interpolated attributes body | expected: [echo-ok] | got: []
```

## Scope


alexskinner:
manual merged and fixed as conflicted

PR #48, opened 2026-05-31
Add CFML test for Application.cfc load errors

Blute:
## Summary

Adds a CFML server-runner test for surfacing `Application.cfc` load/execution errors.

The fixture has an `Application.cfc` that throws immediately and an `index.cfm` body that would output `direct-page-executed` if the application load error were skipped.

## Why this matters

Lucee applications expect Application.cfc load failures to stop the request. Silently falling through to the target page hides configuration/startup failures and can make broken applications appear partially functional.

## Current RustCFML result

When served via RustCFML on current upstream, the new assertions fail:

```text
FAIL: Application.cfc load error returns failure status | expected truthy | got: [false]
FAIL: Application.cfc load error does not execute target page | expected truthy | got: [false]
```

Directly hitting the fixture currently returns:

```text
HTTP/1.1 200 OK

direct-page-executed
```

## Scope

### DIFF
--- a/tests/lifecycle/application_load_error/Application.cfc
+++ b/tests/lifecycle/application_load_error/Application.cfc
@@ -1,0 +1,3 @@
+<cfcomponent>
+    <cfthrow message="broken application load">
+</cfcomponent>
--- a/tests/lifecycle/application_load_error/index.cfm
+++ b/tests/lifecycle/application_load_error/index.cfm
@@ -1,0 +1,1 @@
+<cfoutput>direct-page-executed</cfoutput>
--- a/tests/lifecycle/test_application_load_errors.cfm
+++ b/tests/lifecycle/test_application_load_errors.cfm
@@ -1,0 +1,19 @@
+<cfscript>
+suiteBegin("Lifecycle: Application.cfc load errors");
+
+serverPort = structKeyExists(cgi, "server_port") ? trim(cgi.server_port) : "";
+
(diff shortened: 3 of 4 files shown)


alexskinner:
Thanks @Blute — landed the runtime fix plus your test in 671903e on main.

  Root cause: load_application_cfc ran the Application.cfc pseudo-constructor with let _ = …, discarding any error, and returned None on
  failure — which execute_with_lifecycle treats as "no Application.cfc, fall through to the page." So a throwing Application.cfc was
  indistinguishable from no Application.cfc, and the target page ran with a 200.

  Fix: load_application_cfc now returns Result<Option<CfmlValue>, CfmlError>:
  - Err(e) — pseudo-constructor threw or the file failed to compile → request aborts (server renders 500)
  - Ok(None) — no usable component found → existing fall-through preserved
  - Ok(Some(t)) — loaded normally

  VM program/swap-depth state is restored before the early return so the VM stays consistent.

  Your test (fixtures + runner wiring) is included verbatim, credited to you. Verified:
  - Direct hit on the fixture now returns 500 (was 200 + direct-page-executed)
  - Lifecycle suite over HTTP: 0/2 → 2/2
  - Full CLI suite 3226/3226, cargo test green, and the wasm32-unknown-unknown (cfml-worker) build is clean

  Closing in favour of the merged commit on main.

PR #49, opened 2026-05-31
Add CFML test for front controller fallback

Blute:
## Summary

Adds a CFML server-runner test for configured front-controller fallback routing.

The fixture configures `server.fallback.template` in `tests/.cfconfig.json`, then requests an unresolved URL and expects RustCFML to execute the fallback template while preserving the original path in `url.route` and the original query string.

## Expected Lucee-compatible behaviour

Framework-style CFML apps often rely on the web server to route unresolved paths through a front controller. When RustCFML owns the HTTP server directly, the same behaviour needs to be expressible in RustCFML server config so a request such as `/missing/path?probe=abc` can be handled by a configured CFML template.

## Current RustCFML behaviour

The served test currently returns a `404 Not Found` response for the unresolved path instead of executing the configured fallback template.

## Validation

CLI mode skips the server-only assertion:

```sh
cargo run --quiet -- tests/server/test_front_controller_fallback.cfm
```

Served mode demonstrates the current failure:

```sh
cargo run --quiet -- --serve tests --port 8572
curl -sS http://127.0.0.1:8572/server/test_front_controller_fallback.cfm
```

Observed failure:

```text
FAIL | Server: front controller fallback | 0/2 passed (2 failed)
FAIL: unresolved route is served by fallback template | expected: [fallback|route=/missing/path|probe=abc] | got: [<html><body><h1>404 Not Found</h1><p>The requested URL /missing/path was not found.</p></body></html>]
FAIL: front controller fallback returns success status | expected: [200 OK] | got: [404 Not Found]
```

### DIFF
--- a/tests/.cfconfig.json
+++ b/tests/.cfconfig.json
@@ -2,7 +2,11 @@
   "server": {
     "port": 9999,
     "welcomeFiles": ["index.cfm", "default.cfm"],
-    "maxRequestBodySize": 5242880
+    "maxRequestBodySize": 5242880,
+    "fallback": {
+      "template": "/server/front_controller_fallback.cfm",
+      "routeParam": "route"
+    }
   },
   "runtime": {
     "locale": "en-GB",
--- a/tests/runner.cfm
+++ b/tests/runner.cfm
@@ -143,6 +143,9 @@ try { include "tags/test_tags_whitespace.cfm"; } catch (any e) { writeOutput("ER
 try { include "includes/test_variables_scope_includes.cfm"; } catch (any e) { writeOutput("ERROR | includes/test_variables_scope_includes.cfm | " & e.message & chr(10)); }
 try { include "includes/test_named_args_includes.cfm"; } catch (any e) { writeOutput("ERROR | includes/test_named_args_includes.cfm | " & e.message & chr(10)); }
 
+// --- Server ---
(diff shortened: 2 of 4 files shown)


alexskinner:
Thanks @Blute — landed the server.fallback feature plus your test on main (9ab6ba9), with a follow-up (41e4681) adjusting how the test
  runs.

  Feature: added server.fallback config (template + routeParam). When the built-in web server can't resolve a URL to a file and a
  fallback template is configured, it now runs that template with the original path exposed in url.<routeParam> (default route) and the
  original query string preserved, instead of 404ing. The route value is percent-encoded internally so paths round-trip cleanly through
  the query parser. Implemented in the request handler (crates/cli/src/lib.rs) + schema (crates/cfml-config/src/schema.rs).

  Test: included your fixture + assertions (credited to you), with one change — it's gated behind an explicit ?servertests=1 opt-in.
  This is a server-feature test that needs the server started with a specific doc root + the fallback cfconfig, which conflicts with
  other HTTP-mode suites if run in the default pass (they assume different doc-root/config setups). So it now skips in the normal
  CLI/runner run and executes only when invoked as /tests/runner.cfm?servertests=1 against a suitably-configured server.

  Verified end-to-end: a hit on /missing/path?probe=abc returns 200 with fallback|route=/missing/path|probe=abc; full CLI suite
  3227/3227, cargo test and the wasm32-unknown-unknown build green.

  Closing in favour of the merged commits on main.

PR #51, opened 2026-05-31
Document cached function source metadata issue

Blute:
Adds a markdown note plus a focused Rust integration test describing the VM-internal `source_file` metadata issue found during the Moopa port.

The test compiles a CFC through `compile_file_cached()` and asserts that the compiled function records the originating CFC path in `function.source_file`.

## Validation

Native targeted test currently fails as expected, illustrating the issue:

```sh
cargo test -p cfml-vm --test compiled_function_source_file cached_file_compilation_records_function_source_file
```

Observed failure:

```text
assertion `left == right` failed
  left: Some("/app/lib/db.cfc")
 right: None
```

WASM checks attempted per project expectation:

```sh
cargo build -p cfml-worker -p rustcfml-wasm --target wasm32-unknown-unknown
cargo test -p cfml-vm --target wasm32-unknown-unknown --test compiled_function_source_file --no-run
```

Both could not run in this local environment because the `wasm32-unknown-unknown` target is not installed and `rustup` is not available here to add it:

```text
can't find crate for `core`
the `wasm32-unknown-unknown` target may not be installed
zsh:1: command not found: rustup
```

### DIFF
--- a/crates/cfml-vm/tests/compiled_function_source_file.rs
+++ b/crates/cfml-vm/tests/compiled_function_source_file.rs
@@ -1,0 +1,35 @@
+use cfml_common::vfs::{EmbeddedFs, Vfs};
+use cfml_vm::compile_file_cached;
+use std::collections::HashMap;
+use std::sync::Arc;
+
+const VROOT: &str = "/app";
+
+#[test]
+fn cached_file_compilation_records_function_source_file() {
+    let mut files = HashMap::new();
+    files.insert(
+        "lib/db.cfc".to_string(),
+        r#"
+<cfcomponent>
+    <cffunction name="read">
+        <cfargument name="table_name" />
+        <cfreturn arguments.table_name />
+    </cffunction>
+</cfcomponent>
+"#
+        .as_bytes()
+        .to_vec(),
+    );
+
+    let vfs: Arc<dyn Vfs> = Arc::new(EmbeddedFs::new(files, VROOT.to_string()));
+    let component_path = format!("{}/lib/db.cfc", VROOT);
+    let program = compile_file_cached(&component_path, None, vfs.as_ref()).unwrap();
+    let function = program
+        .functions
+        .iter()
+        .find(|function| function.name.eq_ignore_ascii_case("read"))
+        .expect("compiled read function");
+
+    assert_eq!(Some(component_path.as_str()), function.source_file.as_deref());
+}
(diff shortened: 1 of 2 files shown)


alexskinner:
Thanks for this @Blute, and for the focused repro — you were right that there's a real cross-request bug here. The source_file angle turned out
  not to be the mechanism, but chasing it down led straight to the actual defect, which is now fixed.

  On the source_file hypothesis specifically: I traced every reader of BytecodeFunction.source_file. The cross-request rebind path doesn't use it
  — a stored function dispatches via a volatile per-request index into program.functions, and app-scope functions are kept callable across
  requests by carrying the compiled BytecodeFunction Arcs and remapping the indices (cached_functions + remap_func_indices), not by re-resolving
  from the source file. The field has exactly one consumer today (a stack-trace template string) and it already falls back to the VM's current
  source file. So populating it wouldn't change dispatch/rebind behaviour — which is why I'm not taking the field + assertion as-is.

  But the underlying problem was real. I built a cross-request repro (several app-scope services with shared init/read/save/checkAccess,
  inheritance, struct/array nesting, a factory-returned instance, and a closure) and forced the function table to shift between requests. Most
  shapes held — but a closure / function reference defined in Application.cfc (or the page) and stored in application scope went stale: its
  func_idx sits below the onApplicationStart offset, and the request-end cache refresh was dropping exactly those. On a later request with a
  differently-sized page, the stored index then dangled onto whatever now occupied that slot. That lines up with the cached-app-scope-method
  misbehaviour you hit in the Moopa port.

  Fixed in 8fc6222, released in v0.40.0. The cache now carries every function reachable from application scope (reachability is recomputed each
  request, so overwritten functions still drop out — no stale retention, no unbounded growth). Regression coverage added in
  crates/cfml-vm/tests/xreq_appscope_dispatch.rs.

  Could you re-test Moopa against v0.40.0? If you still see cross-request misdispatch, please open an issue with the smallest app-scope shape
  that reproduces it.

  Also worth saying: your instinct that app-scope functions want a stable identity rather than a volatile per-request index is the right one —
  that's the proper long-term fix and would make this whole bug class impossible. It's on my list as a follow-up (a persistent per-application
  function table / interned identity), and source_file may well become load-bearing there.

  Closing, since the real issue is resolved by a different mechanism — but this was a genuinely useful report, thanks.

PR #52, opened 2026-05-31
Document PostgreSQL UUID parameter issue

Blute:
Adds a markdown note describing the PostgreSQL UUID string parameter compatibility issue from the Moopa port.

### DIFF
--- a/docs/compatibility-notes/postgres-uuid-params.md
+++ b/docs/compatibility-notes/postgres-uuid-params.md
@@ -1,0 +1,56 @@
+# PostgreSQL UUID text parameters
+
+## Summary
+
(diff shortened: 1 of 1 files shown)


bpamiri:
Also relevant to Wheels. Wheels has first-class UUID primary-key support: the migrator emits `uuid` columns (`vendor/wheels/migrator/TableDefinition.cfc`) and the model layer generates string UUIDs for UUID-based keys and binds them as params on create/find (`vendor/wheels/model/create.cfc` — `$isUUIDColumn`/`generateUUID`). So string-UUID → `uuid` column is a path real Wheels apps hit directly on Postgres, and the related QueryColumn case applies too. Indicating interest.

PR #53, opened 2026-05-31
Document PostgreSQL UNKNOWN parameter issue

Blute:
Adds a markdown note describing the PostgreSQL UNKNOWN target parameter encoding issue from the Moopa port.

### DIFF
--- a/docs/compatibility-notes/postgres-unknown-params.md
+++ b/docs/compatibility-notes/postgres-unknown-params.md
@@ -1,0 +1,66 @@
+# PostgreSQL UNKNOWN parameter targets
+
+## Summary
+
+PostgreSQL sometimes asks parameter encoders to encode a value for target type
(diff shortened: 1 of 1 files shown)


bpamiri:
Flagging interest from the Wheels framework port — this would hit Wheels broadly on Postgres. Wheels builds parameterized INSERT/UPDATE/SELECT for essentially every persistence operation through its QueryBuilder + cfqueryparam (`vendor/wheels/model/query/QueryBuilder.cfc`, `properties.cfc`), with the adapter deriving cfsqltype from column metadata (`databaseAdapters/Base.cfc`). Many of those framework-generated statements don't pin a concrete type early, so `Type::UNKNOWN` is exactly the path Wheels' inserts/updates would land on against a Postgres — or CockroachDB, whose adapter extends Postgres — datasource. +1 on encoding UNKNOWN targets as text.

PR #54, opened 2026-05-31
Document PostgreSQL multi-statement mutation issue

Blute:
Adds a markdown note describing the PostgreSQL multi-statement mutation issue from the Moopa port and a suggested fix shape.


bpamiri:
Noting Wheels interest, though narrower for us than the UUID/UNKNOWN param cases. Wheels' migrator `execute()` runs arbitrary user-authored SQL, which is frequently multi-statement (seed data + schema changes), so this would help Wheels migrations against Postgres. Lower priority than the param-encoding issues for our boot path, but +1.

PR #57, opened 2026-06-02
Add cross-engine test for undeclared named arguments keeping their names

bpamiri:
Continuing the Wheels-on-RustCFML work. Calling a function with **more named arguments than it declares** behaves differently on RustCFML, and it blocks the Wheels boot.

## The gap

On Lucee 5/6/7, Adobe CF 2018-2025, and BoxLang, extra (undeclared) named arguments remain reachable **by name** in the `arguments` scope:

```cfml
function f(required string a) {
    // called as f(a="A", b="B", c="C")
    return StructKeyExists(arguments, "b");   // true on JVM engines
}
```

On RustCFML the extras are stored **positionally** — `arguments` ends up with numeric keys (`a,1,2,3,...`) instead of `a,b,c`, so `StructKeyExists(arguments, "b")` is `false` and the value can't be read by name or forwarded.

This underpins a very common framework idiom: declare a few params, accept arbitrary extra named args, and forward the whole `arguments` struct onward.

## Where Wheels hits it

`vendor/wheels/Global.cfc` `$createObjectFromRoot(required string path, required string fileName, required string method)` is called with **extra** named args:

```cfml
$createObjectFromRoot(path="wheels", fileName="Plugins", method="$init",
                      pluginPath=local.pluginPath, deletePluginDirectories=..., ...);
```

It then does `local.argumentCollection = arguments;` and `Invoke(instance, method, argumentCollection)`. On RustCFML the `pluginPath` (and the other extras) aren't keyed by name in `arguments`, so `Plugins.$init` receives an **empty** `pluginPath` → `ExpandPath("")` resolves to the webroot → the plugin-directory scan runs against the wrong path and the boot dies in `directoryList`.

## The test

`tests/core/test_undeclared_named_args.cfm` calls a fixture method declaring only `a` with extra named args `b`/`c`, and asserts they're reachable by name.

| Engine | Result |
|--------|--------|
| Lucee 7.0.2 | `a=A,b=B,c=C` (pass) |
| RustCFML 0.41.0 | `MISSING (keys=a,1,2,3,4,5)` (fail) |

Full suite on 0.41.0: 3201/3202, this the only (intended) red.

### DIFF
--- a/tests/core/UndeclaredArgFixture.cfc
+++ b/tests/core/UndeclaredArgFixture.cfc
@@ -1,0 +1,19 @@
+// Fixture for the "undeclared named arguments keep their names" behavior.
+// `capture()` declares only `a`, but is called with extra named args (b, c).
+// On Lucee/Adobe CF/BoxLang those extras remain reachable by name in the
+// `arguments` scope; on RustCFML they are stored positionally (numeric keys),
+// so StructKeyExists(arguments, "b") is false.
+//
+// Wheels relies on this in vendor/wheels/Global.cfc $createObjectFromRoot: it
+// declares path/fileName/method, receives extra named args (e.g. pluginPath),
+// then forwards the whole `arguments` struct to the target method.
+component {
+	public string function probe() {
+		return capture(a = "A", b = "B", c = "C");
+	}
+	private string function capture(required string a) {
+		return StructKeyExists(arguments, "b") && StructKeyExists(arguments, "c")
+			? "a=" & arguments.a & ",b=" & arguments.b & ",c=" & arguments.c
+			: "MISSING (keys=" & StructKeyList(arguments) & ")";
+	}
+}
--- a/tests/core/test_undeclared_named_args.cfm
+++ b/tests/core/test_undeclared_named_args.cfm
@@ -1,0 +1,30 @@
+<cfscript>
+suiteBegin("Core: undeclared named arguments keep their names");
+
+// ============================================================
+// Background
+// ============================================================
+// Calling a function with MORE named arguments than it declares is standard
+// CFML on Lucee 5/6/7, Adobe CF 2018-2025, and BoxLang: the extra named args
+// remain reachable BY NAME in the `arguments` scope. This underpins the common
+// framework pattern of declaring a few params, accepting arbitrary extra named
+// args, and forwarding `arguments` on to another method.
+//
+// On RustCFML the extra (undeclared) named args are stored POSITIONALLY — the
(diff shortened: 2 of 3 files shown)


bpamiri:
Adding field evidence from porting a real framework (Wheels) on top of this gap, in case it helps prioritise.

This argument-scope behaviour is broader and more severe than the minimal repro suggests:

**1. It also hits functions that declare _no_ formal parameters.** A `function f(){ ... }` called with named args (`f(directory="...", sort="...")`) stores every arg under a numeric key — `arguments.directory` is unreachable. In Wheels this silently broke three core primitives at once: `$directory()`, the universal `$invoke()` dispatcher, and `$createObjectFromRoot()`'s arg-forwarding. Each had to be given explicit formal params to recover the names.

**2. Even functions _with_ declared params get numeric-key pollution.** Calling a declared-param function with named args yields a mixed scope like `action,1,directory,2,sort,3` — the named keys are present, but duplicated positionally. (Standard engines yield just `action,directory,sort`.)

**3. In the full framework it escalates from a clean error to a hard VM panic.** During mixin integration — `Mapper.init()` → `$integrateFunctions(createObject("wheels.Global"))` over a 126-method component, invoked via the DI container as `getInstance(...).init(argumentCollection={...})` — RustCFML 0.50.0 aborts the worker thread:

```
thread 'tokio-runtime-worker' panicked at crates/cfml-vm/src/lib.rs:3082:59:
index out of bounds: the len is 1 but the index is 10
... JoinError::Panic at crates/cli/src/lib.rs:1203
```

The connection is dropped (curl sees HTTP 000). This pipeline underpins every Mapper / Model / Controller, so it is firmly on the critical boot path.

**Honest caveat:** I could *not* minimise the panic to a small standalone repro — a faithful 126-method isolation of the same `getMetaData().functions` copy loop driven through the same `init(argumentCollection=...)` path passes cleanly on RustCFML (and on Lucee). So the panic appears to require the full-framework argument-scope state to trigger, and I can't hand you a one-file reproducer for it yet. But it is a Rust panic (always a bug) and it traces to the same argument-scope-fidelity issue this PR's test covers.

**Net:** with Wheels-side workarounds for the lost names, the boot clears the entire `onApplicationStart` lifecycle on RustCFML and stops only at this panic. A fix that carries named-arg fidelity through nested / `argumentCollection` / no-formal-params invocations would very likely clear all of it at once. Happy to instrument further against a dev build if that would help track down the `lib.rs:3082` index path.

alexskinner:
sorted

PR #61, opened 2026-06-03
Add CFML test for returned service chain

Blute:
## What changed

Adds a CFML regression test and short compatibility note for chained method calls where the second method is invoked on a CFC returned by the first method.

The test covers the Moopa-style shape:

```cfml
application.lib.db.getService("moo_profile").login(...)
```

## Why

Moopa dynamically resolves table services and immediately calls methods on the returned service. Lucee keeps the original factory receiver intact after the chained call.

## Validation

  - observed current upstream regression: `ERROR | oop/test_returned_service_chain.cfm | Component [oop.ChainService] has no function with name [kind]`.
  - this indicates the original factory receiver was clobbered by the returned service during the chain.
- `cargo test -p rustcfml-wasm`
  - passes.

No Rust implementation changes are included.

- `wasm-pack build crates/wasm --target web`
  - not run locally because `wasm-pack` is not installed.

### DIFF
--- a/docs/compatibility-notes/returned-service-chain.md
+++ b/docs/compatibility-notes/returned-service-chain.md
@@ -1,0 +1,15 @@
+# Returned Service Chain Receiver Preservation
+
+Moopa resolves table services dynamically and immediately calls methods on the
+returned service:
+
+```cfml
+application.lib.db.getService("moo_profile").login(...)
+```
+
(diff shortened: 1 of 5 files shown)


alexskinner:
Thanks @Blute! Your three test files landed in ff55f15 (v0.52.0) along with the VM-side identity-guard fix for the chained-writeback
  ▎ clobber. Closing as merged.

Issue #55, opened 2026-06-01
cfhttp body with control flow silently drops cfhttpparams (needs runtime param collection, like cfquery)


alexskinner:
## Summary

  `cfhttpparam`s are collected at **parse time** by a static scan of the `cfhttp`
  body, so any `<cfhttpparam>` / `cfhttpparam(...)` that lives inside a loop or
  conditional never gets built — the control-flow body is never executed to
  produce them. Lucee 5/6/7, Adobe CF 2018-2025, and BoxLang all build the params
  at runtime. This is the same class of problem `cfquery` had before it was fixed
  in 28af97d ("cfquery: build SQL and params at runtime when body has control-flow
  tags").

  Affects **both** the tag form and the cfscript form.

  ## Reproduction (all on v0.40.0)

  **Tag form** — the loop never runs, so the index var is undefined:
  ```cfml
  <cfset coll = ["a","b"]>
  <cfhttp url="http://localhost/" method="get" result="r">
      <cfloop array="#coll#" index="item">
          <cfhttpparam name="#item#" value="1" type="formfield">
      </cfloop>
  </cfhttp>
  <!--- Runtime Error: Variable 'item' is undefined --->

  Script form, for-in in the body — parse error:
  cfhttp(url="http://localhost/", method="get") {
      for (item in coll) {
          cfhttpparam(name=item, value="1", type="formfield");
      }
  }
  <!--- Parse error: Expected RParen, found In --->

  Script form, even with no loop — the paren form isn't recognised as a
  body-tag at all:
  cfhttp(url="http://localhost/", method="get") {
      cfhttpparam(name="a", value="1", type="formfield");
  }
  <!--- Parse error: Expected RBrace, found Semicolon --->

  Root cause

  - Tag form: parse_cfhttpparam_tags(&body) in crates/cfml-compiler/src/tag_parser.rs
  (~line 599) statically extracts <cfhttpparam> tags from the body text; the
  enclosing <cfloop>/<cfif> is not executed.
  - Script attribute form (http url="..." { httpparam ...; }) in
  crates/cfml-compiler/src/parser.rs (~lines 366-425) only recognises
  httpparam statements and skips everything else in the body.
  - Script paren form (cfhttp(...) { ... }) is not in the body-tag list
  (is_body_tag, parser.rs ~line 896), so it parses as a function call followed
  by a bare {} block.

  Proposed fix

  Mirror the cfquery runtime approach (28af97d / body_has_control_flow): when the
[...]

PR #68, opened 2026-06-05
Add cross-engine test for script cfquery statement body

bpamiri:
The script-form `cfquery(...) { ... }` body can contain CFML statements (assignments, loops, conditionals) interleaved with SQL and `cfqueryparam` — the standard idiom for building dynamic queries on Lucee, Adobe CF 2018+, and BoxLang. On v0.54.0 RustCFML parses the body as a struct literal and rejects the first statement:

```
function f() { cfquery(name="q") { local.sql = ""; ... } }
  RustCFML 0.54.0 → Parse error: Expected RBrace, found Semicolon
  Lucee/Adobe/BoxLang → parses fine
```

Verified parse-only on both engines (body kept in a never-called function, so no datasource is needed): RustCFML errors at the first `;`-terminated statement; Lucee parses it. An empty `cfquery(...){}` body parses on RustCFML, so the brace block itself is accepted — it's specifically statement content inside it that's rejected (the body is treated as a struct literal rather than a statement block).

Real-world context: Wheels' database adapter builds *every* query with this form (`databaseAdapters/Base.cfc`), so it blocks DB-backed apps. This is distinct from #55 (which is about runtime `cfhttpparam` collection); here it's the parse of the statement body. The sibling `test_tags_script_syntax_body.cfm` already covers the tags that do parse (`cfsavecontent`/`cflock`/`transaction`); this adds the `cfquery` case. The same struct-literal-vs-statement-block ambiguity also affects script `cfhttp(...) { ... }` bodies.


alexskinner:
done

PR #69, opened 2026-06-05
Add cross-engine test for tags inside a block comment

bpamiri:
A `/* ... */` block comment in a **script-component body** must be opaque: literal CFML tags inside it (`<cfset>`, `<cfoutput>`) are documentation, not markup. Lucee, Adobe CF, and BoxLang ignore the comment interior and compile the component normally. On v0.55.0 RustCFML lexes and **executes** the tags inside the comment (the component file flips to template-echo mode), so the CFC doesn't load as a normal component:

```
component {
  /* <cfset x = 1>  <cfoutput>#x#</cfoutput> */
  function ping() { return "pong"; }
}

createObject("component","BlockCommentTags").ping()
  Lucee / Adobe / BoxLang → "pong"
  RustCFML 0.55.0         → component is mangled (tags in the comment get run); ping() does not return "pong"
```

Verified: reproduces identically on 0.50.0 → 0.55.0 (long-standing, not a recent regression); in plain `<cfscript>` the comment is correctly suppressed — it's specifically the **component-body** parse path. Closed tags only here on purpose (an *unclosed* tag in a comment is a separate hazard). Real-world: Wheels' `Test.cfc` documents `<cfset>`/`<cfoutput>` usage inside a `/* */` doc comment, which makes the whole file echo+execute on RustCFML. Added to the existing comments suite alongside `test_hash_in_comments.cfm`.

### DIFF
--- a/tests/comments/BlockCommentTags.cfc
+++ b/tests/comments/BlockCommentTags.cfc
@@ -1,0 +1,15 @@
+component {
+
+    /*
+     * Documentation example of the tag form — this is a comment, NOT markup.
+     * The literal tags below must be ignored by the lexer; the component must
+     * compile to a normal CFC. (Closed tags only.)
+     *
+     *   <cfset example = 1>
+     *   <cfoutput>#example#</cfoutput>
+     */
+    public string function ping() {
+        return "pong";
+    }
+
+}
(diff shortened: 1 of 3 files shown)


alexskinner:
done

PR #72, opened 2026-06-06
Add CFML test for cflog/cfmail tag-attribute interpolation

Blute:
## What

Adds a cross-engine CFML test pinning that the `<cflog>` and `<cfmail>` **tag** forms interpolate `#...#` segments in a quoted attribute that also contains literal text (e.g. `text="hello #who# end"`).

## Why

The merged tag-attribute interpolation work (#62) applied uniform `#...#` interpolation to `cfthrow`/`cfargument`/`cffile`/`cfparam`, and its test asserts the rule holds "regardless of which tag." But the `<cflog>` and `<cfmail>` tag arms still route attribute values through `strip_hashes`, which only handles a value that is *entirely* one `#expr#`. A mixed literal+interpolation value emits malformed script and fails to **parse**.

The previously merged cflog test (#44) only exercised the CFScript `log`/`cflog()` statement forms, which use a different parse path — so the tag-form gap was never covered.

## Shape

Because this is a hard parse error (not just wrong output as with `cfthrow`), and parse errors escape `try/catch` and would abort the runner, the cases live in runtime-instantiated fixture CFCs — the same pattern as `test_component_declaration_attributes.cfm` / `test_mapping_include.cfm`. Controls (single bare `#expr#`) parse today and guard the fixture wiring; the gap fixtures are **expected to fail on current upstream** until `cflog`/`cfmail` attributes route through `format_attr_value`.

Test-only; no engine change. Motivated by Moopa's `access_policy.cfc` `<cflog ... text="...#...#...ms">`.


alexskinner:
closing and sorting manually

PR #74, opened 2026-06-06
Add CFML test for cfhttp timeout attribute interpolation

Blute:
## What

Adds a cross-engine CFML test pinning that `<cfhttp timeout="#expr#">` interpolates, like every other cfhttp attribute.

## Why

The merged cfhttp interpolation work routes `url`/`method`/`charset`/`username`/`password`/`useragent`/`proxyserver` through the interpolation path, but `timeout` is emitted verbatim into the generated `cfhttp({ ... })` call. So `timeout="#arguments.timeout#"` produces `timeout: #arguments.timeout#` (literal hashes in CFScript) and fails to **parse**. A literal `timeout="60"` works because it has no hashes.

## Shape

Parse-class gap → runtime-instantiated fixtures. The control (literal timeout) parses today and guards the fixture wiring; the interpolated fixture is **expected to fail on current upstream** until `timeout` is interpolated like the other cfhttp attributes. The cfhttp call is guarded by `<cfif false>` so it is compiled but never executed (no network).

Test-only; no engine change. Motivated by Moopa's `cloudflare_stream.cfc` `isPlaybackReady` (`<cfhttp ... timeout="#arguments.timeout#">`).


alexskinner:
don e manually

PR #77, opened 2026-06-07
Add cross-engine test for local scope frame isolation

bpamiri:
`local.X` is per-call scope — identical to `var X`. Each function invocation gets its own `local` scope, so a callee's `local.x = ...` must not alter a caller's `local.x` of the same name. On RustCFML the explicit-prefix form leaks across frames:

```
function callee(){ local.x = 99;          return local.x; }
function caller(){ local.x = "ORIGINAL";  callee();        return local.x; }

caller()
  Lucee / Adobe / BoxLang → "ORIGINAL"   (each frame has its own local)
  RustCFML v0.72.0        → 99           (callee's local.x clobbered the caller's)
```

The boundary is precise (verified on v0.72.0 and v0.65.0 — longstanding, not new in the stable-table rewrite):

- **`var x`** form is correctly frame-isolated → returns `"ORIGINAL"`. ✓
- **distinct** local names (`local.y` vs `local.x`) don't collide → `"ORIGINAL"`. ✓
- only **same-named `local.X`** across caller/callee collides. 🐛

Since `var x` and `local.x` are the same scope in CFML, the two should behave identically; today only `var x` is isolated. The existing `compat_engine/test_scope_behavior.cfm` "Local Scope (LDEV-0112)" suite covers `local.x == var x` *within* one frame — this adds the *across-frame* isolation case it doesn't reach. Test (2) is a `var`-form control that passes on every engine and guards the assertion wiring; tests (1) and (3) are expected to fail on current upstream until `local.X` gets per-frame storage.

Real-world context: this is the last engine-side blocker for booting the Wheels framework. Wheels uses the `local.rv` convention in hundreds of functions; a builder's `local.rv` (the request params struct) is silently overwritten when it calls a helper (`$get()`) that also uses `local.rv`, so dispatch never sees the real params. Renaming isn't feasible at that scale and wouldn't durably hold. Caught via 4-point instrumentation of the real dispatch chain — the params struct was struct-valued through `$createParams` until a `local.obf = $get(...)` call flipped the caller's `local.rv`.

### DIFF
--- a/tests/core/test_local_scope_frame_isolation.cfm
+++ b/tests/core/test_local_scope_frame_isolation.cfm
@@ -1,0 +1,51 @@
+<cfscript>
+suiteBegin("Core: local scope frame isolation");
+
+// `local.X` is per-CALL scope — identical to `var X`. Each function invocation
+// gets its OWN `local` scope, so a callee's `local.x = ...` MUST NOT alter a
+// caller's `local.x` of the same name. Lucee, Adobe CF, and BoxLang all isolate
+// `local` per call frame (and treat `local.x` and `var x` identically).
+//
+// The existing single-frame cases in compat_engine/test_scope_behavior.cfm cover
+// `local.x == var x` WITHIN one frame; these cover isolation ACROSS frames.
+
+// (1) The gap: a callee's same-named local.x leaks into the caller's frame.
+function frameIsoCallee() {
+    local.x = 99;
+    return local.x;
+}
+function frameIsoCaller() {
+    local.x = "ORIGINAL";
+    frameIsoCallee();
+    return local.x;
+}
+assert("callee local.x does not clobber caller local.x", frameIsoCaller(), "ORIGINAL");
+
+// (2) Control: the `var x` form (isolated on every engine) — guards the wiring.
+function frameIsoCalleeVar() {
+    var x = 99;
+    return x;
+}
+function frameIsoCallerVar() {
+    var x = "ORIGINAL";
+    frameIsoCalleeVar();
+    return x;
+}
+assert("callee var x does not clobber caller var x", frameIsoCallerVar(), "ORIGINAL");
+
+// (3) Real-world shape: a helper assigned into the caller must not corrupt the
+// caller's same-named local. Mirrors framework code where a builder uses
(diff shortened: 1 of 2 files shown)


bpamiri:
A sharper characterization from triaging this against the Wheels boot: the leak is specifically in **`local.`-prefixed reference resolution**, independent of how the variable was declared. Cross-engine, RustCFML v0.83.0 vs Lucee:

| # | form | RustCFML | Lucee |
|---|------|----------|-------|
| 1 | caller `local.x="ORIGINAL"`; callee `local.x=99`; return `local.x` | `99` | `ORIGINAL` |
| 2 | `var x="ORIGINAL"`; callee `var x=99`; return **unscoped** `x` | `ORIGINAL` | `ORIGINAL` |
| 3 | `var x="ORIGINAL"`; callee `local.x=99`; return **`local.x`** | `99` | `ORIGINAL` |
| 4 | distinct names (caller `local.x` vs callee `local.y`) | `ORIGINAL` | `ORIGINAL` |
| 5 | `var rv={a:1}; local.rv.b=2`; callee `local.rv="GG"` | clobbered (`rv` == `"GG"`) | `struct[a,b]` |

Takeaways:

- A `var` declaration does **not** protect a value once it is read or written through the `local.` prefix (row 3). So this can't be sidestepped by aliasing `var`↔`local` — the `local` scope itself appears to be shared across call frames rather than allocated per invocation.
- Only `var x` **accessed unscoped** (row 2) and distinct names (row 4) are isolated today.
- Nested/subscript writes — `local.rv.b = …`, `local.rv[k] = …` (row 5) — leak as well, and these have no `var` form in CFML, so struct-building code cannot avoid the `local.` prefix.

Canonical one-liner (the rest follow the same shape):

```
function callee(){ local.x = 99;          return local.x; }
function caller(){ local.x = "ORIGINAL";  callee();        return local.x; }
caller()   // RustCFML: 99   |  Lucee/Adobe/BoxLang: "ORIGINAL"
```

Real-world impact: this is the last engine-side blocker for booting the Wheels framework. `local.rv` is the result-struct convention across hundreds of framework functions, so a builder's `local.rv` is silently overwritten by a callee (e.g. `$get()`) that also uses `local.rv`, and the request params struct never survives to dispatch.

alexskinner:
closed manually

PR #78, opened 2026-06-07
Add CFML test for single hash-expression type preservation

Blute:
## What

Pins that a double-quoted string whose entire content is a single `#expression#` (no surrounding literal text) yields the expression's **native value/type**, matching Lucee/ACF — `"#someStruct#"` is a struct, `"#someArray#"` is an array, not a stringified copy.

## Why

RustCFML currently coerces the single interpolated value to a string in the expression path, so `"#someStruct#"` becomes the lossy text `{k: v}` (`isStruct` false, member access empty). Moopa — and most CFML frameworks — rely on the type-preserving behaviour constantly, e.g. passing a struct through a quoted argument:

```cfml
application.service.moo_file.upload( data = "#request.data#" )   // expects the struct
```

On RustCFML the struct arrives as a string and downstream reads come back empty.

## Coverage

All contexts that exhibit the gap, plus controls:
- **gaps** (expected-fail on current upstream): `cfset` (struct + array, incl. member access / length), named arg, positional arg, struct-literal value, array element, `cfreturn`.
- **controls** (pass today): bare assignment (no quotes), mixed literal+`#expr#` (must stay a string), and `<cfloop array="#arr#">` — the tag-attribute path **already** preserves type via `single_hash_expr`, which is the inconsistency this pins for the expression path.

`cfml-codegen` `compile_expression`, the `Expression::StringInterpolation` arm (`crates/cfml-compiler`→`cfml-codegen/src/compiler.rs`). It always coerces the first part to a string (`push String("")` + `Concat`) even when there is exactly one part. When `interp.parts.len() == 1` and the sole part is the interpolated expression, emit its value without the empty-string `Concat`; keep concatenation for the multi-part (mixed literal + `#expr#`) case.

### DIFF
--- a/tests/core/test_single_hash_type_preservation.cfm
+++ b/tests/core/test_single_hash_type_preservation.cfm
@@ -1,0 +1,105 @@
+<cfscript>
+suiteBegin("Core: single hash-expression preserves type");
+</cfscript>
+
+<!---
+    ============================================================
+    Background
+    ============================================================
+    On Lucee and Adobe ColdFusion, a double-quoted string whose ENTIRE content
+    is a single #expression# (no literal text before, between, or after) yields
+    the expression's NATIVE VALUE/TYPE -- it is NOT coerced to a string:
+
+        st = { a: 1 };
+        x  = "#st#";       // x is the STRUCT, isStruct(x) is true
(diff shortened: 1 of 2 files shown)


alexskinner:
Close manually

PR #79, opened 2026-06-08
Add CFML test for component methods named like builtins

Blute:
## What

Pins that a **component method** whose name collides with a built-in function (`canonicalize`, `reverse`, …) registers and dispatches via `obj.method()` — as on Lucee/ACF, where object method dispatch takes precedence over the builtin.

## Why — regression

`cc6279a` ("Lucee-parity reject of builtin redefinition", v0.75.0) added a correct rule: a **top-level** user function cannot redefine a builtin (`function abs(){}` throws on Lucee). But the guard in the `DefineFunction` op (`crates/cfml-vm/src/lib.rs`) fires for **any** non-`__` function named like a builtin — including **component methods**. So a CFC method named `canonicalize` never registers, and worse, it **poisons the whole component** (sibling ordinary methods stop dispatching too).

This regressed real apps: Moopa's `route_url.cfc` defines `canonicalize()` and the entire routing layer calls `routeUrl.canonicalize(url)` — every routed request 500s with *"Component [route_url] has no function with name [canonicalize]"*. Renaming isn't viable (any of ~hundreds of builtin names could collide).

`tests/core/test_builtin_shadowing.cfm` (added in the same commit) only covers top-level UDF redefinition and local closures — it never exercises component methods, which is the gap.

## Coverage

- `canonicalize` and `reverse` component methods must dispatch (expected-fail today).
- a builtin-named method must **not** poison a sibling ordinary method on the same component (expected-fail today — shows the blast radius).
- control: an ordinary method dispatches (passes today).

## What to fix

`crates/cfml-vm/src/lib.rs`, the `DefineFunction` op handler: scope the *"The name [X] is already used by a built in Function"* guard to **top-level** function declarations only — a function defined as a component method must register and be dispatchable via `obj.name()` even when its name matches a builtin. Top-level redefinition should still throw (existing `test_builtin_shadowing.cfm` stays green).

### DIFF
--- a/tests/oop/BuiltinNameMethodFixture.cfc
+++ b/tests/oop/BuiltinNameMethodFixture.cfc
@@ -1,0 +1,29 @@
+<!---
+    Fixture: a component that defines methods whose names collide with built-in
+    functions (canonicalize, reverse) alongside an ordinary method. On
+    Lucee/ACF this is legal -- a method named like a builtin is callable via
+    obj.method(); object method dispatch takes precedence over the builtin.
+
+    On RustCFML v0.75.0+ (commit cc6279a, "Lucee-parity reject of builtin
+    redefinition") the DefineFunction op throws / drops any non-"__" function
+    whose name is a builtin, which over-reaches to COMPONENT METHODS -- so these
+    methods do not register and obj.canonicalize()/obj.reverse() fail with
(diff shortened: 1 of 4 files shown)


alexskinner:
Manually close

PR #80, opened 2026-06-09
test(stdlib): ListRest should return the literal remainder (Lucee parity)

bpamiri:
## Cross-engine gap: `ListRest` should return the literal remainder

On Lucee 5/6/7, Adobe ColdFusion, and BoxLang, `ListRest` returns the **literal substring from element 2 onward** — preserving interior/trailing empty elements *and* the original delimiter characters from the source string. RustCFML 0.92.0 instead collapses empty elements and normalizes the delimiter set to a single canonical char.

| call | RustCFML 0.92.0 | Lucee (correct) |
|------|-----------------|-----------------|
| `listRest(Reverse("/home/index"), "/")` | `"emoh"` | `"emoh/"` |
| `listRest("a/b/", "/")` | `"b"` | `"b/"` |
| `listRest("a/b;c/", "/;")` | `"b/c"` | `"b;c/"` |
| `listRest("a,,b,,c")` | `"b,c"` | `"b,,c"` |

Sibling list fns (`ListFirst`/`ListLast`/`ListLen`) correctly **keep** their empty-collapsing semantics and agree on both engines — the test asserts those as controls so it can't pass for the wrong reason.

### Why it matters
Surfaced while booting the Wheels framework on RustCFML: view-path resolution does `Reverse(path)` then `ListRest(...,"/")` to peel a segment, and the dropped trailing delimiter corrupts the reconstructed path. It's a general correctness gap, not Wheels-specific.

### Suggested fix shape (stdlib)
`ListRest(list, delims)` = the original string starting at the first character of element 2: skip any leading delimiter run, skip element 1, skip the single delimiter run that follows it, then return the remainder of the **original** string verbatim (no re-join, no delimiter rewrite). That naturally preserves interior/trailing empties and mixed delimiter chars. (`includeEmptyFields`/4th-arg behavior unaffected — this is the default 2/3-arg path.)

### Verification
- **RustCFML 0.92.0:** 4 divergent assertions FAIL, 2 baseline + 3 sibling controls PASS.
- **Lucee 7:** all 9 PASS.

Test registered in the stdlib list cluster of `tests/runner.cfm`.

### DIFF
--- a/tests/runner.cfm
+++ b/tests/runner.cfm
@@ -70,6 +70,7 @@ try { include "stdlib/test_struct_higher_order.cfm"; } catch (any e) { writeOutp
 try { include "stdlib/test_math_functions.cfm"; } catch (any e) { writeOutput("ERROR | stdlib/test_math_functions.cfm | " & e.message & chr(10)); }
 try { include "stdlib/test_date_functions.cfm"; } catch (any e) { writeOutput("ERROR | stdlib/test_date_functions.cfm | " & e.message & chr(10)); }
 try { include "stdlib/test_list_functions.cfm"; } catch (any e) { writeOutput("ERROR | stdlib/test_list_functions.cfm | " & e.message & chr(10)); }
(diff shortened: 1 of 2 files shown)


alexskinner:
Manually merged and closed credit for the test provided

PR #81, opened 2026-06-09
test(core): single-token string interpolation should preserve the value type

bpamiri:
## Cross-engine gap: single-token string interpolation should preserve the value type

On Lucee 5/6/7, Adobe ColdFusion, and BoxLang, a quoted string whose **entire** content is a single interpolation token — `"#expr#"` — is a special case: it does **not** stringify the value, it yields the value *itself* with its type intact. So `a = "#someStruct#"` leaves `a` a struct; `a = "#someArray#"` leaves `a` an array.

RustCFML 0.92.0 stringifies even this single-token form:

| expression | RustCFML 0.92.0 | Lucee (correct) |
|------------|-----------------|-----------------|
| `s = {x:1}; isStruct("#s#")` | `false` | `true` |
| `a = [1,2]; isArray("#a#")` | `false` | `true` |

As soon as the literal has any *other* part (a leading/trailing char, or a second token) it degrades to ordinary concatenation on every engine — asserted here as controls (`"#1##2#"` → `"12"`, `"#1#-#2#"` → `"1-2"`, `"x#n#"` → `"x42"`), so the test can't pass for the wrong reason.

### Why it matters
Surfaced booting the Wheels framework: the view layer round-trips single-token interpolations through its `contentFor`/section plumbing, and the unexpected stringification corrupts struct/array payloads. General language-semantics gap, not Wheels-specific.

### Suggested fix shape (codegen)
When an interpolated-string node has exactly one part and no surrounding literal text, emit that part's expression **directly** (preserving its runtime type) instead of wrapping it in `String("") & <expr>`. Multi-part nodes keep the concat path.

### Verification
- **RustCFML 0.92.0:** the 2 struct/array assertions FAIL; the numeric check + 4 controls PASS (5/7).
- **Lucee 7:** all 7 PASS.

### DIFF
--- a/tests/core/test_single_part_interpolation_preserves_type.cfm
+++ b/tests/core/test_single_part_interpolation_preserves_type.cfm
@@ -1,0 +1,54 @@
+<cfscript>
+suiteBegin("Core: single-part string interpolation preserves the value type");
+
+// ============================================================
+// Background
+// ============================================================
+// On Lucee 5/6/7, Adobe ColdFusion, and BoxLang, a quoted string whose
+// ENTIRE content is a single interpolation token -- "#expr#" -- is a
+// special case: it does NOT stringify the value, it yields the VALUE
+// ITSELF with its type preserved. So `a = "#someStruct#"` leaves `a`
+// a struct, `a = "#someArray#"` leaves `a` an array, etc.
+//
+// As soon as the literal has ANY other part (a leading/trailing
+// character, or a second token) it degrades to ordinary string
+// concatenation/stringification on every engine -- that is the CONTROL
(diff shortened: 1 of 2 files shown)


bpamiri:
Closing as a duplicate — this exact gap (single-hash `"#expr#"` interpolation not preserving the native value type) was fixed in **v0.92.1** via #78, with the same root cause and fix (short-circuiting the `parts.len() == 1` case in the StringInterpolation codegen arm). Verified on the v0.92.1 binary: `s={x:1}; a="#s#"; isStruct(a)` now returns `true`. Thanks!

PR #82, opened 2026-06-09
test(core): named-arg call to a declared-param fn should leak no numeric alias key

bpamiri:
## Cross-engine gap: named-arg call to a declared-param fn leaks a numeric alias key

A named-argument call to a function that **declares** parameters must populate the `arguments` scope by name only. RustCFML 0.92.0 leaks a spurious purely-numeric positional-alias key into the scope when a named argument lands in (or alongside) a declared positional slot.

```cfml
function cf(any position="last", any overwrite="false") { return StructKeyList(arguments); }
cf(body="X", overwrite=true)
```

| engine | `StructKeyList(arguments)` |
|--------|-----------------------------|
| RustCFML 0.92.0 | `"overwrite,2,body,position"` ← spurious `"2"` |
| Lucee 7 | `"position,overwrite,body"` |

The junk numeric key inflates `StructCount`, makes a `for (k in arguments)` loop see a bogus `"2"`, and round-trips through any argument-forwarding helper.

### Why it matters
Surfaced booting the Wheels framework: `contentFor()` detects which section to write by reading `StructKeyList(arguments)` / `arguments` keys, and the phantom `"2"` derails section selection — the layout body comes out empty. Wheels' option-forwarding helpers do this scope-introspection pervasively.

The **paramless** case (`g(){...}` called `g(alpha=1, sort=2)`) is already correct on both engines and is asserted here as a control, isolating the regression to the declared-positional-slot path.

### Verification
- **RustCFML 0.92.0:** the 3 numeric-key assertions FAIL; the 4 name-presence checks + paramless control + 2 more PASS (7/10).
- **Lucee:** all 10 PASS.

Assertions are presence/absence only — Lucee uppercases and reorders the `arguments` scope, so no order-sensitive check is used.

### DIFF
--- a/tests/core/test_named_args_no_numeric_alias.cfm
+++ b/tests/core/test_named_args_no_numeric_alias.cfm
@@ -1,0 +1,76 @@
+<cfscript>
+suiteBegin("Core: named-arg call to a declared-param fn leaks no numeric alias key");
+
+// A named-argument call to a function that DECLARES parameters must populate
+// the arguments scope by NAME only. RustCFML 0.92.0 leaks a spurious numeric
+// positional-alias key into the arguments scope when a named argument lands in
+// a declared positional slot:
+//
+//   function cf(any position="last", any overwrite="false"){ return StructKeyList(arguments); }
+//   cf(body="X", overwrite=true)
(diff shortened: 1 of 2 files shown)


alexskinner:
manually fixed

PR #84, opened 2026-06-09
test(core): typed return type on a top-level cfscript function (Lucee parity)

bpamiri:
## Cross-engine gap: typed return type on a top-level cfscript function

On Lucee 5/6/7, Adobe ColdFusion, and BoxLang, a **top-level** (non-component) cfscript function may carry a return-type annotation — `struct function f()` — exactly like a component method. RustCFML 0.92.0 misparses the leading type token at the top level as a bare *expression statement*, so the declaration throws at runtime:

```cfml
struct function f(){ return {a:1}; }   // RustCFML: Runtime Error: Variable 'struct' is undefined
```

Every type keyword regresses identically — `struct`, `array`, `string`, `numeric`, `boolean`, `query`, `any`, `void`.

### Scope of the regression (the two controls in the test)
| context | RustCFML 0.92.0 |
|---------|-----------------|
| `struct function f()` at top level of a `.cfm` | ❌ throws |
| `struct function f()` **inside a component** | ✅ works |
| **untyped** `function f()` at top level | ✅ works |

So the parser handles the type annotation for component methods but not for top-level function declarations.

### Why it matters
Surfaced booting the Wheels framework. `vendor/wheels/public/helpers.cfm:293` declares:

```cfml
struct function $returnInternalDocumentation(required array documentScope, required array ignore) { ... }
```

and many view-helper `.cfm` files declare typed top-level functions. This is general CFML — typed top-level functions are common — not Wheels-specific.

### Verification
- **Lucee:** 4/4 pass (control + struct/array/string typed top-level fns).
- **RustCFML 0.92.0:** runtime error at the first typed declaration. **Longstanding** — fails identically on 0.65.0 and 0.92.0.

On RustCFML this manifests as a **runtime** error (caught by the runner's `try/catch` → `ERROR | ... | Variable 'struct' is undefined`), so it registers safely in `tests/runner.cfm` without aborting the suite.

### DIFF
--- a/tests/core/test_typed_toplevel_function_return_type.cfm
+++ b/tests/core/test_typed_toplevel_function_return_type.cfm
@@ -1,0 +1,38 @@
+<cfscript>
+suiteBegin("Core: typed return type on a top-level cfscript function");
+
+// On Lucee 5/6/7, Adobe ColdFusion, and BoxLang, a top-level (non-component)
+// cfscript function may carry a return-type annotation -- `struct function f()`
+// -- exactly like a component method. RustCFML 0.92.0 misparses the leading
+// type token at the top level as a bare expression STATEMENT, so
+// `struct function f(){...}` throws "Variable 'struct' is undefined" at runtime.
+//
+// The same declaration INSIDE a component works, and an UNTYPED top-level
+// function works -- both are exercised here as controls so the test can't pass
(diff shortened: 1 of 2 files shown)


alexskinner:
fixed manually

PR #85, opened 2026-06-10
test(types): query["col"][row] bracket-column cell access should return the cell (Lucee parity)

bpamiri:
## Cross-engine gap: `query["col"][row]` bracket-column cell access returns empty

`query["columnName"][rowNumber]` (bracket-column + index) must return the cell value, exactly like `query.columnName[rowNumber]` (dot form). On Lucee 5/6/7, Adobe ColdFusion, and BoxLang both forms work. RustCFML returns an **empty** value for the bracket form, while the dot form works.

```cfml
q = queryNew("column_name,type_name", "varchar,varchar", [["id","INTEGER"],["title","TEXT"]]);

q.column_name[1]       // "id"   on both engines
q["column_name"][1]    // ""     on RustCFML   /   "id" on Lucee
q["column_name"][2]    // ""     on RustCFML   /   "title" on Lucee
```

### Why it matters
Frameworks read query cells via the **bracket** form whenever the column name is dynamic (held in a variable). Wheels' ORM column processing does exactly this:

```cfml
// vendor/wheels/Model.cfc
local.columnName = local.columns["column_name"][local.i];
```

With the bracket form returning blanks, every introspected column comes back empty → the model can't build its property map → no model works. This is the current wall for running the Wheels ORM on RustCFML.

### Verification
- **RustCFML 0.98.0:** the 4 bracket-form assertions FAIL (`got: []`); the 2 dot-form controls PASS (2/6).
- **Lucee:** all 5 pass.

The test uses `queryNew()` (no datasource) so it's portable and deterministic. Dot-form controls guard the wiring. Verified red across 0.90.x→0.98.0; dot form has always worked.

### DIFF
--- a/tests/runner.cfm
+++ b/tests/runner.cfm
@@ -54,6 +54,7 @@ try { include "types/test_nested_writeback.cfm"; } catch (any e) { writeOutput("
 try { include "types/test_query.cfm"; } catch (any e) { writeOutput("ERROR | types/test_query.cfm | " & e.message & chr(10)); }
 try { include "types/test_query_column.cfm"; } catch (any e) { writeOutput("ERROR | types/test_query_column.cfm | " & e.message & chr(10)); }
 try { include "types/test_query_reference.cfm"; } catch (any e) { writeOutput("ERROR | types/test_query_reference.cfm | " & e.message & chr(10)); }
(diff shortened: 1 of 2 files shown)


alexskinner:
merged a fix thanks

PR #86, opened 2026-06-10
Add CFML test for null as a bare struct-literal key

Blute:
## What

Pins that a bare `null` directly before a key separator (`:` or `=`) in a struct literal names the **key** `"null"` — matching Lucee/ACF — rather than being parsed as the null value literal.

## Why

RustCFML currently parses `{ value: "", null: true }` with an **empty-string key** in place of `null`, so the entry is unreachable by name and `structKeyExists(s, "null")` is false.

This silently breaks `cfqueryparam`: the tag compiles its attributes into exactly this struct shape (`{value: ..., cfsqltype: ..., null: true}`), so `null="true"` vanishes and the param binds `""` instead of SQL NULL. Found running Moopa on RustCFML: the route-registry sync inserts a route endpoint with `created_by` NULL when no session user exists — the lost flag bound `""` into a uuid column and aborted application startup.

## Coverage

- **gaps** (expected-fail on current upstream): bare `null` key with `:` separator (key exists + value is true), and with `=` separator (the cfqueryparam attribute style).
- **controls** (pass today): quoted `"null"` key; a plain literal without keyword keys.

`crates/cfml-compiler/src/parser.rs`, `parse_struct_literal`: when the current token is a keyword literal (`null` — and `true`/`false` deserve the same treatment) and the next token is `Colon`/`Equal`, consume it as a string-literal KEY instead of letting `parse_expression` produce the value literal.


alexskinner:
merged a fix

PR #87, opened 2026-06-10
Add CFML test for cfqueryparam-style queryExecute param structs

Blute:
## What

Pins that a cfqueryparam-style struct (`{value, cfsqltype, null, ...}`) is accepted wherever a `queryExecute` parameter is expected — positional array entries and named-struct entries alike — and binds the struct's **value** (or SQL NULL when `null=true`), matching Lucee. The `cfquery` tag's `null="true"` is covered end-to-end since the tag compiles to exactly this struct form.

## Why

RustCFML currently binds the **stringified struct** for named params: on sqlite the lookup matches nothing (0 rows); on postgres the prepared statement fails with `error serializing parameter 0` (the text `{value: ..., cfsqltype: ...}` can't bind to a typed column). Found running Moopa on RustCFML: every `db.save` INSERT goes through this form, so the first new route endpoint after an upgrade aborted application startup.

A diagnostic note for the postgres path: the surfaced message is just `error serializing parameter 0` — the underlying cause (`cannot bind "..." as PostgreSQL uuid: ...`) is in the error's `source()` chain and currently dropped, which makes this gap expensive to trace. Chaining it would help independent of the fix.

## Coverage

Uses the `sqlite://` tempfile + graceful-skip pattern from `test_cfqueryparam_interpolated_value.cfm` (skips on engines without the sqlite scheme, e.g. Lucee under CommandBox).

- **controls** (pass today): positional param struct (the array path already flattens these); `null=true` via a *quoted* `"null"` key.
- **gaps** (expected-fail on current upstream): named param struct (binds 0 rows instead of the value, 2 asserts); `null=true` via the *bare* `null` key (overlaps the struct-literal null-key test on purpose — `cfquery`'s `null="true"` only works when both hold); `<cfqueryparam null="true">` end-to-end through the `cfquery` tag.

## Where the fixes go (for reference; not included here)

- postgres: `crates/cfml-stdlib/src/pg_sql.rs`, `prepare_pg_statements` — unwrap the param struct per entry in both the `Array` and `Struct` branches (honouring `null`), instead of `query_column_scalar().clone()` on the raw struct.
- sqlite: `crates/cfml-stdlib/src/builtins.rs`, `build_sqlite_params` named branch — same unwrap before `cfml_to_sqlite` (the positional path already flattens via the shared pre-processing; mysql's named branch likely needs the same).

### DIFF
--- a/tests/runner.cfm
+++ b/tests/runner.cfm
@@ -157,6 +157,7 @@ try { include "tags/test_tags_cfmail.cfm"; } catch (any e) { writeOutput("ERROR 
 try { include "tags/test_tags_cfcache.cfm"; } catch (any e) { writeOutput("ERROR | tags/test_tags_cfcache.cfm | " & e.message & chr(10)); }
 try { include "tags/test_tags_cfstoredproc.cfm"; } catch (any e) { writeOutput("ERROR | tags/test_tags_cfstoredproc.cfm | " & e.message & chr(10)); }
 try { include "tags/test_tags_cfqueryparam_attribute_collection.cfm"; } catch (any e) { writeOutput("ERROR | tags/test_tags_cfqueryparam_attribute_collection.cfm | " & e.message & chr(10)); }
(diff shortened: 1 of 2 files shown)


alexskinner:
merged a fix

PR #89, opened 2026-06-10
test(core): invoke() should spread an argumentCollection key in its argument struct

bpamiri:
## Cross-engine gap: invoke() drops an `argumentCollection` key in its argument struct

`argumentCollection` is a special key on every CFML argument-passing surface. When the argument struct handed to the positional BIF `invoke(objectOrName, method, argStruct)` contains an `argumentCollection` key, Lucee spreads the **inner** struct into the callee's `arguments` scope — the same contract as `fn(argumentCollection = st)` and cfinvoke's `argumentCollection` attribute. RustCFML does not special-case the key: the inner struct is silently dropped (it neither spreads nor lands as a literal named argument), so declared params keep their defaults.

```cfml
function target(string name = "DEFAULT", string path = "P0") {
    return "name=" & arguments.name & " path=" & arguments.path;
}
invoke(o, "target", { argumentCollection = { name = "posts" } })
```

| engine | result |
|--------|--------|
| RustCFML 0.101.0 | `"name=DEFAULT path=P0"` ← inner struct dropped |
| Lucee 5.4.8.2 | `"name=posts path=P0"` (spread; unnamed param keeps its declared default) |

The plain 3-arg struct form `invoke(o, "target", { name = "posts", path = "/x" })` already agrees on both engines and is asserted as a control.

### What the test asserts

- **CONTROL:** the plain 3-arg struct form binds declared params (passes on both engines — wiring guard).
- A nested `argumentCollection` key is **spread**: the declared param it names receives the inner value, the param it does not name keeps its declared default.
- An inner struct naming every param binds them all.
- The spread also applies when invoking by component NAME (string first argument).
- The spread populates the callee's `arguments` scope even for a method with no declared params (presence-keyed via `listFindNoCase`, never order- or case-sensitive).
- The literal key `"argumentCollection"` itself must NOT leak into the callee — pins the fix shape (spread the inner struct, don't bind the key by name). Already true on both engines.

### Why it matters

Wheels rides this contract on every request. `$doubleCheckedLock()` forwards its `executeArgs` as `$invoke(method = ..., argumentCollection = executeArgs)`, and `$invoke()` pushes that struct through the engine's dynamic-invoke machinery. Controller class resolution (`$controller` → `$doubleCheckedLock` → `$createControllerClass`) runs on this path — an engine that drops the key hands the resolver EMPTY arguments, and every user controller action silently becomes a no-op. This was the single gap that made all user controller actions no-ops when booting the Wheels blog tutorial app.

This is a runtime gap (wrong values, no parse error), so the test is registered in `tests/runner.cfm`, placed immediately after `test_invoke_canonical_forms.cfm`. The new fixture `tests/core/InvokeArgCollTarget.cfc` lives next to the test.

### Verification

- **RustCFML 0.101.0** (release binary, full `tests/runner.cfm`): suite completes 3614/3618 across 435 suites — the only 4 failures are this test's spread assertions (`got: [name=DEFAULT path=P0]` / empty arguments scope); the control + literal-key guard pass.
- **Lucee 5.4.8.2** (CommandBox): all 6 assertions PASS — `SUMMARY: 6/6 passed / ALL TESTS PASSED`.

### DIFF
--- a/tests/core/InvokeArgCollTarget.cfc
+++ b/tests/core/InvokeArgCollTarget.cfc
@@ -1,0 +1,16 @@
+// Target for the invoke() argumentCollection-spread test.
+component {
+	public string function target(string name = "DEFAULT", string path = "P0") {
+		return "name=" & arguments.name & " path=" & arguments.path;
+	}
+
+	// Reports which keys actually landed in the arguments scope.
+	public string function argKeyList() {
+		return structKeyList(arguments);
+	}
+
+	// True only if the literal key "argumentCollection" leaked through unspread.
+	public boolean function hasLiteralArgumentCollectionKey() {
+		return structKeyExists(arguments, "argumentCollection");
+	}
+}
--- a/tests/core/test_invoke_argumentcollection_spread.cfm
+++ b/tests/core/test_invoke_argumentcollection_spread.cfm
@@ -1,0 +1,72 @@
+<cfscript>
+suiteBegin("Core: invoke() spreads an `argumentCollection` key in its argument struct");
+
+// ============================================================
+// Background  (companion to test_invoke_canonical_forms.cfm)
+// ============================================================
+// `argumentCollection` is a SPECIAL key on every CFML argument-passing
+// surface. When the argument struct handed to the positional BIF
+// invoke(objectOrName, method, argStruct) contains an `argumentCollection`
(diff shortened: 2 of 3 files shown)


alexskinner:
sorted manually merged

PR #91, opened 2026-06-10
test(core): cross-case struct write should update the key in place, not fork a second key

bpamiri:
## Cross-engine gap: a differently-cased struct WRITE forks the key instead of updating in place

CFML structs are case-insensitive containers. RustCFML already falls back across case on READS (`u.zip` finds a key created as `u["ZIP"]`), but a write through a different casing of an existing key creates a **second physical key** — the same logical key stored twice with two independent values:

```cfml
s = {a: 1};
s["A"] = 99;
```

| engine | `structCount(s)` | `structKeyList(s)` | `s.a` | `s["A"]` |
|--------|------------------|--------------------|-------|----------|
| RustCFML 0.105.0 | **2** | `"a,A"` | **1** | 99 |
| Lucee 5.4.8.2 | 1 | one key | 99 | 99 |

All three write paths fork, not just bracket-onto-literal:

```cfml
t = structNew();
t["zip"] = "30303";
t.ZIP = "90210";
// RustCFML: structCount=2, keys "zip,ZIP", t["zip"]="30303" (stale)
// Lucee:    structCount=1, keys "zip" (first-written casing preserved), every casing reads "90210"
```

Delete coherence breaks too: after a cross-case write, `structDelete(d, "name")` on RustCFML removes only the exactly-matching key and leaves the forked `NAME` behind; on Lucee the struct is empty.

Target behavior (the stated goal: **case preservation by default, Lucee parity always**): one key — the first-written casing wins the key list, and later cross-case writes update the value in place without re-casing the stored key.

### What the test asserts

- `structCount == 1` and a single-element `structKeyList` after a cross-case write, on all three write paths (bracket→literal key, dot→bracket key, dot→literal key)
- the last-written value is readable through every casing (dot, bracket original-case, bracket write-case)
- first-written casing is preserved **exactly** in `structKeyList` — asserted only for a bracket-created key. Lucee 5.x uppercases unquoted struct-literal / dot keys at creation (`{count:1}` stores `COUNT`), so an exact-casing assert on a literal-created key would be a false failure on a conforming engine; those keys are asserted case-insensitively instead.
- `structDelete` under the original casing after a cross-case write leaves the struct EMPTY
- CONTROLS (pass on both engines): cross-case read fallback, case-insensitive `structKeyExists`, same-case write round-trip, case-insensitive `structDelete`

### Why it matters

Set-one-case / read-another-case is everywhere in Wheels: url/form params merged into `params`, query `columnList` lookups, option/config struct merging (`structAppend` over caller-supplied option structs). With a forked key the read-side case-fallback finds the **stale** first key and silently returns the old value — no error, wrong data, the worst failure mode for a framework boot.

### Verification

- **RustCFML 0.105.0** (release binary, full `tests/runner.cfm` on this branch): suite completes 3621/3634 across 435 suites — the only 13 failures are this test's cross-case-write assertions (forked counts/keys/stale values as shown above); the 7 controls + 2 casing-matched-read consistency pins PASS. Identical 9/22 result with `RUSTCFML_JIT=0` and `RUSTCFML_JIT_THRESHOLD=1` — interpreter semantics, not JIT-specific.
- **Lucee 5.4.8.2** (CommandBox): all 22 PASS.
- Runtime gap (wrong values, no parse error) → registered in the core block of `tests/runner.cfm`.

### DIFF
--- a/tests/core/test_struct_key_case_parity.cfm
+++ b/tests/core/test_struct_key_case_parity.cfm
@@ -1,0 +1,96 @@
+<cfscript>
+suiteBegin("Core: struct keys are case-insensitive on WRITE — cross-case writes update in place");
+
+// CFML structs are case-insensitive containers: a write through ANY casing of
+// an existing key must update that key IN PLACE. RustCFML 0.103.0 falls back
+// across case on READS (s.zip finds a key created as s["ZIP"]), but a
+// differently-cased WRITE forks the key — the same logical key ends up stored
+// twice with two independent values:
+//
+//   s = {a: 1}; s["A"] = 99;
+//     RustCFML 0.103.0 -> structCount=2, keys "a,A", s.a=1, s["A"]=99
+//     Lucee 5.4 / 7    -> structCount=1, every casing reads 99
+//
+// Target behavior (maintainer-stated: case preservation by default, Lucee
+// parity always): ONE key — the first-written casing wins the key list, and
+// later cross-case writes update the value without re-casing the stored key.
+//
+// The fork poisons any set-one-case / read-another-case flow: url/form params,
+// query columnList lookups, option/config struct merging — Wheels does all of
+// these constantly, so a forked key silently reads back a stale value.
+//
+// Note on key-CASING asserts: Lucee 5.x UPPERCASES unquoted struct-literal /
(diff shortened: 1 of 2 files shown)


alexskinner:
closed manually fixed

PR #93, opened 2026-06-10
test(core): undeclared callee must get a fresh, empty local — caller's local invisible to absence-checks/reads

bpamiri:
## Gap

`local` is per-call: every invocation gets a fresh, **empty** local scope. v0.92.0 (PR #77) fixed the declaration side — a callee's `local.x = ...` no longer clobbers the caller — but the **read side still leaks**: a callee that never declares `local.rv` can still see the caller's `local.rv` via `StructKeyExists(local, "rv")`, `isNull(local.rv)`, and plain reads.

```cfm
function innerFn() { return StructKeyExists(local, "rv"); }
function outerFn() { local.rv = false; return innerFn(); }
writeOutput(outerFn());
```

| Engine | `outerFn()` | `isNull(local.rv)` in callee |
|---|---|---|
| RustCFML 0.105.0 | `true` (and `local.rv` reads the caller's `false`) | `false` |
| Lucee 5.4.8.2 | `false` | `true` |

Same result template-level and component-method-level, and in all three execution modes (default, `RUSTCFML_JIT=0`, `RUSTCFML_JIT_THRESHOLD=1`).

## What the test asserts

`tests/core/test_local_scope_absence_leak.cfm` + fixture `tests/core/AbsenceLeakFixture.cfc`:

1. **Template-level**: an undeclared callee sees `StructKeyExists(local, "rv") == false`, `isNull(local.rv) == true`, and cannot read any caller value through `local.rv`.
2. **The exact Wheels default-true tail** — `if (!StructKeyExists(local, "rv")) { local.rv = true; } return local.rv;` — returns `true` even when the caller holds `local.rv = false`.
3. **CONTROL (green on both engines)**: a callee that *declares* `local.rv` first owns its own slot and the caller's value survives — pins the already-fixed #77 contract (`test_local_scope_frame_isolation.cfm`) so the declaration and read contracts can't be conflated.
4. **Component-method shape**: the same probes as methods on one CFC — exactly how Wheels mixins run.

On RustCFML 0.105.0: 1/9 pass (only the CONTROL). On Lucee: 9/9.

## Why it matters for Wheels

This was the silent ORM-save killer in the Wheels boot ladder. Wheels' `$callback()` (`vendor/wheels/model/callbacks.cfc`) ends with the default-true tail above, and the same `StructKeyExists(local, "rv")` absence-check appears in 8 framework files. `$callback()`'s caller also keeps a `local.rv` accumulator (the framework's universal return-value convention), so on RustCFML the tail inherited the caller's `local.rv = false` through the leak: every model callback chain "returned false", and `save()` aborted before issuing its INSERT — no error, no INSERT, nothing to debug from. Fixing the absence-check/read side completes per-frame `local` and unblocks the model callback chain (and every other default-true fallback in the framework).

## Verification

- **RED** — RustCFML 0.105.0 (release binary): `1/9 passed (8 failed)`; identical under default JIT, `RUSTCFML_JIT=0`, and `RUSTCFML_JIT_THRESHOLD=1`. The 8 failures are exactly the absence-check/read assertions; the declares-first CONTROL passes, confirming the v0.92.0 declaration fix is intact.
- **GREEN** — Lucee 5.4.8.2 (CommandBox): `9/9 passed`.
- The gap is runtime-level (wrong values, no parse error), so the test is registered in `tests/runner.cfm`.
- Full `tests/runner.cfm` on this branch (RustCFML 0.105.0): **3613/3621 across 435 suites** — the only 8 failures are this suite's absence-check/read assertions; no abort.

### DIFF
--- a/tests/core/AbsenceLeakFixture.cfc
+++ b/tests/core/AbsenceLeakFixture.cfc
@@ -1,0 +1,34 @@
+component {
+
+	// Undeclared callee: must get a fresh, empty `local` regardless of the
+	// caller's same-named local.rv.
+	public struct function innerProbe() {
+		var r = {
+			seesRv   = StructKeyExists(local, "rv"),
+			rvIsNull = isNull(local.rv),
+			leaked   = ""
+		};
+		if (r.seesRv) r.leaked = toString(local.rv);
+		return r;
+	}
+
+	public struct function outerProbe() {
+		local.rv = false;
+		return innerProbe();
+	}
+
+	// The Wheels $callback() default-true tail: returns true unless THIS
+	// body set local.rv (it never does here).
+	public boolean function callbackTail() {
+		if (!StructKeyExists(local, "rv")) {
+			local.rv = true;
+		}
+		return local.rv;
+	}
+
+	public boolean function invokeWithAccumulator() {
+		local.rv = false; // caller's own working flag, same conventional name
+		return callbackTail();
+	}
+
+}
--- a/tests/core/test_local_scope_absence_leak.cfm
+++ b/tests/core/test_local_scope_absence_leak.cfm
@@ -1,0 +1,96 @@
+<cfscript>
+suiteBegin("Core: caller's local is invisible to an undeclared callee (absence checks)");
+
+// `local` is per-CALL: every invocation gets a fresh, EMPTY local scope. A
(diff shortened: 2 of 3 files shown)


alexskinner:
closed manually

PR #94, opened 2026-06-10
test(core): bare call to a UDF named 'new' dispatches; method call on a null receiver throws

bpamiri:
## Two composing silent-failure gaps

Both are runtime-level (wrong values / silent nulls — no parse errors), so runner registration is safe.

### (C1) Bare call to a UDF named `new` evaluates to null instead of dispatching

`new` is a soft keyword on Lucee/Adobe CF/BoxLang: with no component path after it, `new (argumentCollection = {...})` is a plain sibling function call, not the new-operator. RustCFML 0.105.0 evaluates the bare call to NULL — the UDF never runs (likely parsed as the new-operator with nothing to instantiate). The explicit `this.new(...)` form already works on both engines (that contract landed with `NewMethodFixture.cfc`); this gap is specifically the bare, unqualified call.

```cfml
component {
    public any function new(struct properties = {}) {
        return { marker = "NEW-RAN", props = arguments.properties };
    }
    public any function createViaBare() {
        return new (argumentCollection = {properties = {a = 1}});
    }
}
```

| expression | RustCFML 0.105.0 | Lucee 5.4.8.2 |
|---|---|---|
| `o.createViaBare()` | `null` — the UDF never runs | struct, `marker = "NEW-RAN"` |
| `o.createViaThis()` (`this.new(...)`) | struct, `marker = "NEW-RAN"` | struct, `marker = "NEW-RAN"` |

### (C2) Method call on a null receiver silently returns null

| expression | RustCFML 0.105.0 | Lucee 5.4.8.2 |
|---|---|---|
| `r = <null-returning call>; r.save(a = 1)` | silently evaluates to `null` | throws |

## What the test asserts

`tests/core/test_new_udf_dispatch_and_null_call.cfm` + fixture `tests/core/NewUdfFixture.cfc`:

1. **(C1)** `createViaBare()` returns a non-null struct whose `marker` proves the `new()` UDF executed, and the `argumentCollection` payload arrived (`props.a == 1`).
2. **(C1 control)** `createViaThis()` dispatches to the same UDF — green on both engines, isolating the gap to the bare form.
3. **(C2)** `.save(a = 1)` on a null receiver (produced by a return-nothing function, independent of C1) must throw — try/catch with a uniquely-named state struct so the flag survives the catch on every engine.
4. **(C2 control)** the same `.save(a = 1)` on a real object works — green on both engines.

## Why it matters for Wheels

This is exactly how the composed gaps bit while laddering Wheels' ORM to a full MVC round trip. Wheels model creation is this composition, verbatim (`vendor/wheels/model/create.cfc:32`):

```cfml
local.rv = new (argumentCollection = arguments);  // (C1) null on RustCFML — the new() UDF never ran
local.rv.save(...);                               // (C2) silent no-op on the null receiver
return local.rv;                                  // create() reports success; no row was ever written
```

On RustCFML, `model("Post").create(...)` returned "success" with zero rows saved and zero errors raised. Either gap alone would at least be loud — C1 with throwing null-calls would crash immediately at `.save()`; C2 with a working bare `new` never sees a null receiver. Composed, they produce the worst failure mode: silent fake success.

## Verification

- **RustCFML 0.105.0** (release binary): suite FAILS as intended — 2/6, only the two controls pass; the four gap assertions fail (`bare new() returned a value`, `marker dispatch`, `argumentCollection forwarded`, `method call on a null receiver throws`). Identical under `RUSTCFML_JIT=0` and `RUSTCFML_JIT_THRESHOLD=1`.
- **Lucee 5.4.8.2** (CommandBox 6.3.2): 6/6 PASS (harness assert helpers renamed `chk*` locally for the CommandBox run because it shadows `assert`; test logic identical).
- Full `tests/runner.cfm` on this branch (RustCFML 0.105.0): **3614/3618 across 435 suites** — the only 4 failures are this suite's gap assertions; no abort.

### DIFF
--- a/tests/core/NewUdfFixture.cfc
+++ b/tests/core/NewUdfFixture.cfc
@@ -1,0 +1,43 @@
+// Fixture for tests/core/test_new_udf_dispatch_and_null_call.cfm.
+//
+// `new` is a SOFT keyword on Lucee/Adobe CF/BoxLang: it introduces the
+// `new Foo()` operator, but it is equally legal as a function name, and a
+// BARE sibling call to it — `new (argumentCollection = {...})` with no
+// component path after the keyword — dispatches to the UDF. Wheels' model
+// creation depends on exactly this shape (vendor/wheels/model/create.cfc:32):
+//
+//     local.rv = new (argumentCollection = arguments);
+//     local.rv.save(...);
+component {
+
+	// The sibling method literally named `new` (the Wheels model shape).
+	public any function new(struct properties = {}) {
+		return {
+			marker = "NEW-RAN",
+			props  = arguments.properties
+		};
+	}
+
+	// The Wheels create() shape: a BARE call to the sibling UDF named `new`,
+	// forwarding arguments via argumentCollection.
+	public any function createViaBare() {
+		return new (argumentCollection = {properties = {a = 1}});
+	}
+
+	// Control: the same call, explicitly this-qualified (works on both engines).
+	public any function createViaThis() {
+		return this.new(argumentCollection = {properties = {a = 1}});
+	}
+
+	// Null source for the second half of the test: a function that returns
+	// nothing yields a null receiver, independent of the bare-new gap above.
+	public any function giveNothing() {
+		return;
+	}
+
+	// Real receiver for the method-call-on-null control.
+	public string function save(numeric a = 0) {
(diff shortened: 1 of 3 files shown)


alexskinner:
manually merged and fixed

PR #100, opened 2026-06-11
test(core): this.method() dot-calls must not detach the frame's this binding from later writes

bpamiri:
## Gap

Inside a component method, a `this.`-DOT-qualified method call (`this.noop()`) must dispatch against the live object and leave the frame's `this` binding alone — Lucee (and Adobe CF, BoxLang) deliver every subsequent `this.key = value` write in that frame, and in any frame it calls, to the caller's object. RustCFML 0.108.0 instead **detaches the frame's `this` binding onto a data-complete SHALLOW COPY** at the moment of the dot-call: every later this-write in the frame (and in frames called after it, any call shape) lands on the detached copy — visible in-frame, **discarded when the detaching frame returns**.

```cfml
component {  // fixture
	public void function noop() {}
	public void function combo() {
		this.noop();      // <- detaches `this` on RustCFML
		this.mk = "X";    // lands on the detached copy
	}
}

o = createObject("component", "Fixture");
o.combo();
structKeyExists(o, "mk");  // Lucee: true | RustCFML 0.108.0: false
```

| After the method returns | Lucee 5.4.8.2 | RustCFML 0.108.0 |
|---|---|---|
| `this.noop(); this.mk = "X"` (new key) | `mk` present | `mk` missing — write discarded |
| `this.noop(); this.pre = "B"` (existing key) | `pre == "B"` | reverts to the pre-call value |
| `this.noop(); bareWrite()` (write in a frame called after the dot-call) | mark present | discarded with the detaching frame |
| `noop()` bare call, then write (control) | survives | survives — no detach |
| `this["noop"]()` bracket call, then write (control) | survives | survives — no detach |
| `f = this.noop` dot-READ, then write (control) | survives | survives — no detach |
| `other.noop()` / `this.member.noop()` (receiver is not `this`), then write (control) | survives | survives — no detach |
| `variables.x = ...` after a dot-call (control) | survives | survives — `variables` never snapshotted |
| `this.bag.n = 7` nested-struct mutation after a dot-call (pin) | escapes | escapes — the copy is shallow, struct refs shared |

The six green controls are the sharp wedge for the fix: the trigger is exactly the member-call shape `this.method()` — bare dispatch, bracket dispatch on `this`, dot-reads, and dot-calls whose receiver is another object all leave `this` attached. The detached copy is data-complete (in-frame reads see the writes; the suite asserts in-frame visibility separately from after-return survival) and shallow (nested-struct mutations escape — pinned so a fix can't silently regress reference semantics for nested containers). The failure is JIT-independent (identical under `RUSTCFML_JIT=0` and `RUSTCFML_JIT_THRESHOLD=1`) and depth-independent (bare recursion to depth 200 followed by a this-write stays green — re-verified on 0.108.0). It also does not propagate upward: caller writes after a poisoned callee returns are fine, and a fresh method call on the same object sees true state — which is exactly what made it hard to localize from above.

## What the test asserts

- Dot-call then NEW-KEY this-write: visible in-frame (green on both engines), and `this.mkNew` survives the return. *(after-return assert red on RustCFML)*
- Dot-call then EXISTING-KEY overwrite: visible in-frame (green on both), and `this.pre == "B"` after return — RustCFML silently reverts it. *(red on RustCFML)*
- Dot-call poisons frames called AFTER it: a this-write made inside a nested BARE call is also discarded on the outer return. *(red on RustCFML)*
- CONTROL (green on both engines): bare call `noop()` does not detach.
- CONTROL (green on both engines): bracket call `this["noop"]()` does not detach.
- CONTROL (green on both engines): dot-READ `f = this.noop` does not detach.
- CONTROL (green on both engines): dot-call on a DIFFERENT object — both `other.noop()` and `this.helperObj.noop()` — does not detach.
- CONTROL (green on both engines): a `variables`-scope write after a dot-call survives.
- PIN (green on both engines): a nested-struct mutation through `this` escapes the detach — the copy is shallow and shares struct references.

## Why it matters for Wheels

This is the gap that broke Wheels model persistence end-to-end while laddering the blog app through full CRUD. The save chain crosses TWO masked detach points:

- `invokeWithTransaction()` calls `this.$hashedConnectionArgs()` (`vendor/wheels/model/transactions.cfc:24`) before dispatching `$save` — so the primary key, timestamps, and persisted-state flags written below it all landed on a detached copy. User-visible: after `update()` the DB row is correct but the caller's object keeps a stale `updatedAt` and reports `hasChanged() = true` with `changedProperties() = [updatedAt]`.
- `$create()` calls `this.columnDataForProperty()` (`vendor/wheels/model/create.cfc:289`) right before assigning the generated primary key — the PK write died one return later, so `create()` handed back an object with no id.

The two points overlapped and mimicked a single strange "transaction boundary" until each was isolated by elimination. Flipping each call site from `this.$x()` to the bare `$x()` — one line per site — restored full Lucee semantics: the sharpest possible demonstration that member-call dispatch clones the receiver while bare and bracket dispatch don't. Any CFC that calls one of its own methods with `this.` qualification and then mutates `this` is affected, so this reaches well beyond the two Wheels sites.

## Verification

- **RustCFML 0.108.0** (release binary): suite fails 3/13 — exactly the three after-return assertions; all three in-frame asserts, all six controls, and the shallow-copy pin pass. Identical results with `RUSTCFML_JIT=0` and with `RUSTCFML_JIT_THRESHOLD=1`.
- **Lucee 5.4.8.2** (CommandBox task runner; same assertions transliterated to a `chk()` harness because CommandBox shadows `assert`): 13/13 pass.
- Full `tests/runner.cfm` on this branch (RustCFML 0.108.0): **3642/3645 across 438 suites** — the only 3 failures are this suite's after-return assertions; no abort.

### DIFF
--- a/tests/core/ThisDotCallDetachFixture.cfc
+++ b/tests/core/ThisDotCallDetachFixture.cfc
@@ -1,0 +1,92 @@
+component {
+
+	// Pseudo-constructor: seed the state the matrix rows assert against.
+	this.pre = "A";
+	this.bag = { n = 0 };
+	variables.vMark = "";
+
+	public void function noop() {
+	}
+
+	public void function writeNested() {
+		this.nestedMark = "N";
+	}
+
+	// ------------------------------------------------------------
+	// RED rows on RustCFML 0.108.0 -- each performs a `this.`-DOT
+	// qualified method call, then a this-write. The write is visible
+	// in-frame (returned string) but discarded when the frame returns.
+	// ------------------------------------------------------------
+
+	public string function dotCallThenNewKeyWrite() {
+		this.noop();        // dot-qualified call: detaches `this` on RustCFML
+		this.mkNew = "X";   // lands on the detached copy
+		return "inFrame=" & structKeyExists(this, "mkNew");
+	}
+
+	public string function dotCallThenOverwrite() {
+		this.noop();
+		this.pre = "B";     // existing key: silently reverts to "A" on return
+		return "inFrame=" & this.pre;
+	}
+
+	public string function dotCallThenNestedBareWrite() {
+		this.noop();
+		writeNested();      // a frame called AFTER the dot-call is poisoned too
+		return "inFrame=" & structKeyExists(this, "nestedMark");
+	}
+
+	// ------------------------------------------------------------
+	// GREEN control rows -- the closest non-detaching neighbours of the
+	// trigger shape. All already pass on RustCFML 0.108.0.
+	// ------------------------------------------------------------
+
+	public string function bareCallThenWrite() {
+		noop();             // bare call: no detach
+		this.mkBare = "X";
+		return "inFrame=" & structKeyExists(this, "mkBare");
+	}
+
+	public string function bracketCallThenWrite() {
+		this["noop"]();     // bracket-dispatch on this: no detach
+		this.mkBracket = "X";
+		return "inFrame=" & structKeyExists(this, "mkBracket");
+	}
+
+	public string function dotReadThenWrite() {
+		var f = this.noop;  // dot-READ (no call): no detach
+		this.mkRead = "X";
+		return "inFrame=" & structKeyExists(this, "mkRead");
+	}
+
+	public string function otherObjectDotCallThenWrite() {
+		var other = createObject("component", "ThisDotCallDetachFixture");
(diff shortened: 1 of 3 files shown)


alexskinner:
closed manually

PR #105, opened 2026-06-11
Add CFML test for cfdirectory recurse through symlinked directories

Blute:
## What

A test that `<cfdirectory action="list" recurse="true">` traverses *symlinked* directories the way Lucee does. POSIX-only: the symlink is created with `/bin/ln -s` via cfexecute, and the test self-skips when the link can't be created (e.g. Windows).

## Why

Lucee's recursive directory listing follows directory symlinks; RustCFML's does not, so files reachable only through a symlinked directory silently vanish from the listing — no error, just missing rows. Real-world hit: an app that assembles its route registry from a recursive `cfdirectory` scan lost every route under a symlinked directory, surfacing as unexplained 404s.

On current main (v0.112.0):

```
FAIL | cfdirectory recurse=true traverses symlinked directories | 2/3 passed (1 failed)
       FAIL: recursive listing traverses the symlinked directory | expected: [2] | got: [1]
```

## Coverage

- fixture: `base/real_dir/Probe.cfc` plus `base/linked_dir → real_dir`
- control: listing the symlinked directory *as the root* finds the file (passes today)
- control: non-recursive listing of the parent sees both entries (passes today)
- gap: `recurse="true" filter="*.cfc"` from the parent must find the file twice (Lucee: 2, current RustCFML: 1)

## Where the fix goes

The recursive walker behind `cfdirectory action="list"`: it should descend into directory symlinks. Worth pairing with cycle protection (track visited canonical paths) since following symlinks makes loops possible — Lucee follows them regardless.


alexskinner:
fix submitted and merged thanks for the PR

PR #106, opened 2026-06-11
Add CFML test for invoke() keeping undeclared named arguments

Blute:
## What

A pure-CFML test (no db, no network) that `invoke()` and `<cfinvoke>` keep named arguments that match no declared `cfargument` — the same behaviour direct method calls already have (covered by `core/test_undeclared_named_args.cfm`).

## Why

In Lucee, undeclared named arguments arrive in the callee's `arguments` scope by name whether the call is direct, via `invoke()`, or via `<cfinvoke>`. RustCFML keeps them on direct calls but DROPS them on the invoke path, so only declared parameters arrive. This silently breaks dynamic-dispatch frameworks: `invoke(handler, endpoint, args)` loses every key the handler didn't declare — in our app, every dynamic route parameter vanished ("No ID provided" on every parameterised route).

On current main (v0.112.0):

```
FAIL | Core: invoke() keeps undeclared named arguments | 2/4 passed (2 failed)
       FAIL: invoke() keeps undeclared named args | expected: [a=A,b=B,c=C] | got: [MISSING (keys=a)]
       FAIL: cfinvoke keeps undeclared named args | expected: [a=A,b=B,c=C] | got: [MISSING (keys=a)]
```

## Coverage

- control: direct call `o.probe(a="A", b="B", c="C")` keeps all three (passes today)
- control: `invoke()` with only the declared argument dispatches fine (passes today)
- gap: `invoke(o, "probe", {a, b, c})` must deliver `b`/`c` in the arguments scope
- gap: `<cfinvoke component="#o#" method="probe" a="A" b="B" c="C">` likewise

## Where the fix goes

The named-argument binding path used by `invoke()`/`__cfinvoke`: after matching declared `cfargument`s, unmatched named arguments should be appended to the callee's arguments scope by name — identical to what the direct named-call path already does.

### DIFF
--- a/tests/core/InvokeUndeclaredArgFixture.cfc
+++ b/tests/core/InvokeUndeclaredArgFixture.cfc
@@ -1,0 +1,10 @@
+component {
+	public string function probe(required string a) {
+		var hasB = StructKeyExists(arguments, "b");
+		var hasC = StructKeyExists(arguments, "c");
+		if (hasB && hasC) {
+			return "a=" & arguments.a & ",b=" & arguments.b & ",c=" & arguments.c;
+		}
+		return "MISSING (keys=" & StructKeyList(arguments) & ")";
+	}
+}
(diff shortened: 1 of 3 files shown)


alexskinner:

PR #110, opened 2026-06-12
test(tags): cfdirectory attributeCollection must deliver the name= result variable

bpamiri:
## What

A self-contained test (scratch dir created and deleted at runtime — no fixtures, no db, no network) that `cfdirectory(attributeCollection = attrs)` delivers the listing query to the variable named by the `name=` attribute, exactly like the direct named-attribute form already does.

```cfml
local.rv = "";
local.args = {action: "list", directory: dir, name: "rv"};
cfdirectory(attributeCollection = "#local.args#");
// Lucee -> local.rv is the listing query
```

| Engine | Result |
|---|---|
| Lucee 5.4.8.2 | `local.rv` is the listing query — 4/4 pass |
| RustCFML 0.130.0, CLI mode | the attribute struct never reaches the tag: throws `cfdirectory: directory not found: ` (empty path) |
| RustCFML 0.130.0, `--serve` mode | attributes get through and the listing runs, but the query is bound to NO scope: `local.rv` stays `""`, `variables.rv` unset, `IsDefined("rv")` false |

The direct form `cfdirectory(action="list", ..., name="rv")` works on RustCFML in both modes (it is the green control in this test), and `cfquery(attributeCollection=...)` delivers its `name=`/`result=` correctly (covered by `tags/test_cfquery_result_delivery.cfm`) — the result-variable write-back plumbing exists; only cfdirectory's attributeCollection path is broken. History: the function-call form itself was #56's territory; the attributeCollection shape regressed separately (CLI throws, serve silently no-delivers).

## Why (the Wheels story)

This is the first thing that kills a Wheels boot on RustCFML — it fires inside `onApplicationStart` before a single request is served:

- `Global.cfc::$directory()` is byte-for-byte the shape under test: copy `arguments` into a plain struct, `cfdirectory(attributeCollection = "#local.args#")`, return the pre-initialized `local.rv`.
- `Plugins.cfc::$folders()` feeds that return value straight into a query-of-queries — the silent `""` surfaces as `Query of Queries: table 'query' not found`, which is exactly what assertion (4) reproduces.
- The same helper backs `$fileExistsNoCase()` → `$objectFileName()`, i.e. EVERY controller and model class-file resolution goes through this path. There is no booting around it.

## Coverage

- control: direct named attrs deliver into a pre-initialized `local.rv` (passes on 0.130.0 today, both modes)
- gap: `attributeCollection = local.args` (struct value)
- gap: the exact Wheels `$directory` shape — arguments copied into a plain struct, string-interpolated `attributeCollection = "#local.args#"`, result read back from `local.rv`
- gap: downstream QoQ over the delivered listing (the `Plugins.cfc::$folders` shape that dies first at boot)

Every cfdirectory call is wrapped in try/catch so the CLI-mode throw registers as an informative assertion failure instead of aborting the bundle — the failure is runtime-level in both modes, so the test is runner-safe.

## Verification

RED — RustCFML 0.130.0, CLI mode (identical under default, `RUSTCFML_JIT=0`, and `RUSTCFML_JIT_THRESHOLD=1`):

```
FAIL | Tags: cfdirectory attributeCollection delivers name= result | 1/4 passed (3 failed)
       FAIL: attributeCollection = struct: name= delivers the listing query | expected: [QUERY rows=2] | got: [ERROR: cfdirectory: directory not found: ]
       FAIL: Wheels $directory shape (interpolated attributeCollection): returns the query | expected: [QUERY rows=2] | got: [NOT-DELIVERED [ERROR: cfdirectory: directory not found: ]]
       FAIL: downstream QoQ over the delivered listing (Plugins.cfc $folders shape) | expected: [rows=2 first=alpha.txt] | got: [ERROR: Query of Queries: table 'src' not found (no such query variable)]
```

RED — RustCFML 0.130.0, `--serve` mode (the silent no-delivery shape):

```
FAIL | Tags: cfdirectory attributeCollection delivers name= result | 1/4 passed (3 failed)
       FAIL: attributeCollection = struct: name= delivers the listing query | expected: [QUERY rows=2] | got: [NOT-DELIVERED []]
       FAIL: Wheels $directory shape (interpolated attributeCollection): returns the query | expected: [QUERY rows=2] | got: [NOT-DELIVERED []]
       FAIL: downstream QoQ over the delivered listing (Plugins.cfc $folders shape) | expected: [rows=2 first=alpha.txt] | got: [ERROR: Query of Queries: table 'src' not found (no such query variable)]
```

GREEN — Lucee 5.4.8.2 (CommandBox task runner; `assert` renamed to `chk` locally because CommandBox shadows `assert` — file content in this PR is the repo-style original):

```
PASS | Tags: cfdirectory attributeCollection delivers name= result | 4/4 passed
```

## Where the fix goes

Two layers of the same attribute-materialization path:

1. CLI mode: attributeCollection expansion for cfdirectory never happens — the struct is dropped before attribute resolution, so `directory` arrives empty and the tag throws.
2. serve mode: attributes are applied but the `name=` result variable is never written back to the caller's frame. cfquery's attributeCollection path already does this write-back correctly (including into function-local scope), so the same mechanism should be wired into cfdirectory. Worth sweeping the other result-producing tags that accept attributeCollection while there: `cffile(attributeCollection = {action: "read", file: f, variable: "rv"})` currently throws `Variable 'cffile' is undefined`, suggesting the same family of gaps.

Full `tests/runner.cfm` on this branch (RustCFML 0.130.0): **3835/3838 across 451 suites** — the only 3 failures are this suite's gap assertions; no abort.

### DIFF
--- a/tests/runner.cfm
+++ b/tests/runner.cfm
@@ -219,6 +219,14 @@ try { include "tags/test_cfdbinfo.cfm"; } catch (any e) { writeOutput("ERROR | t
 try { include "tags/test_cfdirectory_mapping_path.cfm"; } catch (any e) { writeOutput("ERROR | tags/test_cfdirectory_mapping_path.cfm | " & e.message & chr(10)); }
 try { include "tags/test_cfdirectory_function_form.cfm"; } catch (any e) { writeOutput("ERROR | tags/test_cfdirectory_function_form.cfm | " & e.message & chr(10)); }
 try { include "tags/test_cfdirectory_recurse_symlink.cfm"; } catch (any e) { writeOutput("ERROR | tags/test_cfdirectory_recurse_symlink.cfm | " & e.message & chr(10)); }
+// cfdirectory attributeCollection must deliver the name= listing query the
+// same way the direct named-attribute form does (Wheels Global.cfc::$directory
+// boots through exactly this shape; Plugins.cfc::$folders QoQs the result at
+// onApplicationStart). 0.130.0 fails in both modes: CLI drops the attribute
+// struct entirely (throws "cfdirectory: directory not found: " with an empty
+// path); serve mode runs the listing but binds the query to no scope.
+// Runtime-level in both modes (the test catches the CLI throw) -> runner-safe.
+try { include "tags/test_cfdirectory_attrcoll_name.cfm"; } catch (any e) { writeOutput("ERROR | tags/test_cfdirectory_attrcoll_name.cfm | " & e.message & chr(10)); }
 try { include "tags/test_cfhttpparam_runtime_body.cfm"; } catch (any e) { writeOutput("ERROR | tags/test_cfhttpparam_runtime_body.cfm | " & e.message & chr(10)); }
 try { include "tags/test_cfmail_runtime_body.cfm"; } catch (any e) { writeOutput("ERROR | tags/test_cfmail_runtime_body.cfm | " & e.message & chr(10)); }
 try { include "tags/test_cfstoredproc_runtime_body.cfm"; } catch (any e) { writeOutput("ERROR | tags/test_cfstoredproc_runtime_body.cfm | " & e.message & chr(10)); }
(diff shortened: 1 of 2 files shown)


alexskinner:
this is sorted thanks for the test

PR #111, opened 2026-06-12
test(core): nested write on a scope-qualified undeclared variable must auto-vivify a struct

bpamiri:
## Cross-engine gap: nested write on a SCOPE-QUALIFIED undeclared variable does not auto-vivify

In CFML, assigning to a member path of an undeclared variable auto-creates the container as a struct even when the lvalue is scope-qualified. RustCFML fixed the UNSCOPED form (`x.a = 1`, PR #21 / `test_undefined_var_autovivify.cfm`) and the subscript form (`test_subscript_autovivify.cfm`); the scope-qualified flavor is the residual — and on 0.130.0 it fails in two different ways:

```cfml
variables.zzc.name = "x";   // variables.zzc never initialized
request.ctx.user   = "x";   // request.ctx never initialized
local.cache.hit    = "x";   // (inside a function) local.cache never initialized
```

| shape | Lucee 5.4.8.2 | RustCFML 0.130.0 |
|---|---|---|
| `variables.X.name = "v"` (template top level) | vivifies struct, readback OK | **throws** `Variable 'X' is undefined` |
| `request.X.name = "v"` (template top level) | vivifies struct, readback OK | **silent loss**: key registered, `isStruct()` → `false`, readback `""` |
| `local.X.name = "v"` (inside a function) | vivifies struct, readback OK | **silent loss** (same signature) |
| `variables.X.name = "v"` (function / CFC method) | vivifies struct, shared-scope visible | **silent loss** (same signature) |
| `variables.X.a.b = "v"` (CFC method, two missing levels) | vivifies the whole chain | **silent loss** (outer level not a struct) |

The silent flavor is the dangerous one: no error at the write, no error at the read. `structKeyExists(scope, "X")` even returns `true`, but the key is bound to a non-struct empty value, the written member is gone, and reading it back yields `""` instead of throwing — nothing points at the assignment that was dropped.

### What the test asserts

- `variables.X.name = ...` on an undeclared `X` at template level vivifies a struct: readback OK, no exception, `isStruct()` true
- same contract for `request.X.name = ...` at template level
- `local.X.name = ...` inside a function vivifies `local.X` as a struct
- `variables.X.name = ...` inside a template-level function lands in the SHARED variables scope, visible after the call
- CFC method (fixture `ScopedAutoVivFixture.cfc`): `variables.$class.name = "vivified"` with a `$`-prefixed, never-pre-seeded key — the exact Wheels `$initControllerClass` shape; the fixture returns `THREW: ...` / `NO-CONTAINER` / `NOT-A-STRUCT` / `name=[...]` so a failing engine self-diagnoses its flavor
- CFC method: two-level chain `variables.X.a.b = ...` must vivify EVERY missing level (the csrf-mixin chain shape)
- CONTROL (passes on RustCFML today): the same nested write on a PRE-INITIALIZED scoped container — guards the wiring

### Why it matters for Wheels

Controller class initialization opens with exactly this shape (`vendor/wheels/Controller.cfc`, `$initControllerClass()`):

```cfml
variables.$class.name = arguments.name;   // variables.$class never pre-seeded
variables.$class.path = arguments.path;
variables.$class.verifications = [];
variables.$class.filters = [];
variables.$class.cachableActions = [];
```

On RustCFML every one of those writes evaporates silently — the controller "initializes" with no name, no filters, no verifications, no layouts, and dispatch fails far downstream of the real cause (the csrf and caching mixins then read `variables.$class.csrf.type` / `variables.$class.cachableActions` off the broken container). Model class init rides the same contract. This gap was masked by a local workaround since before the test-PR strategy existed; with the cfinvoke family fixed in 0.130.0 it is the next silent killer on the Wheels dispatch path.

### Verification

- **RustCFML 0.130.0** (release binary), staged repo layout (`harness.cfm` + this test + fixture): **1/11** — only the pre-initialized CONTROL passes; all 10 gap assertions fail, no abort. Identical 1/11 with `RUSTCFML_JIT=0` and with `RUSTCFML_JIT_THRESHOLD=1` — interpreter semantics, not JIT-specific.
- **Lucee 5.4.8.2** (CommandBox): all **11/11** PASS. (Adobe CF was not run for this PR; scoped auto-vivification is long-standing shared CFML behavior, and the unscoped sibling test documents the same contract family — but these specific shapes were pinned against Lucee only.)
- Full `tests/runner.cfm` on this branch (RustCFML 0.130.0): **3835/3845 across 451 suites** — baseline on stock main is **3834/3834 across 450 suites**, so the only 10 failures are this suite's gap assertions; no abort.
- Runtime gap (the one throwing case is wrapped in try/catch; the silent flavor needs no wrapper) → registered in `tests/runner.cfm` directly after `core/test_undefined_var_autovivify.cfm`, its unscoped sibling.

### DIFF
--- a/tests/core/ScopedAutoVivFixture.cfc
+++ b/tests/core/ScopedAutoVivFixture.cfc
@@ -1,0 +1,44 @@
+// Fixture for the scope-qualified nested auto-vivification test. Each probe
+// performs the write inside a CFC method (the Wheels $initControllerClass
+// context) and reports exactly what state the variables scope ended up in,
+// so a silent-loss engine fails with a diagnosis instead of a bare mismatch.
+component {
+
+	// The exact Wheels Controller.cfc shape: variables.$class.name = ...
+	// where variables.$class is never pre-seeded ($-prefixed key included).
+	public string function vivClassName() {
+		try {
+			variables.$zzscvivClass.name = "vivified";
+		} catch (any e) {
+			return "THREW: " & e.message;
+		}
+		if (!StructKeyExists(variables, "$zzscvivClass")) {
+			return "NO-CONTAINER";
+		}
+		if (!IsStruct(variables.$zzscvivClass)) {
+			return "NOT-A-STRUCT";
+		}
+		return "name=[" & variables.$zzscvivClass.name & "]";
+	}
+
+	// Two-level chain (csrf mixin shape: variables.$class.csrf.type = ...):
+	// every missing level must vivify as a struct.
+	public string function vivDeep() {
+		try {
+			variables.scvivDeep.a.b = "deep-viv";
+		} catch (any e) {
+			return "THREW: " & e.message;
+		}
+		if (!StructKeyExists(variables, "scvivDeep")) {
+			return "NO-CONTAINER";
+		}
+		if (!IsStruct(variables.scvivDeep)) {
+			return "OUTER-NOT-A-STRUCT";
+		}
+		if (!StructKeyExists(variables.scvivDeep, "a") || !IsStruct(variables.scvivDeep.a)) {
+			return "INNER-NOT-A-STRUCT";
+		}
+		return "b=[" & variables.scvivDeep.a.b & "]";
+	}
+
+}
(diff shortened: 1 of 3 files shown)


alexskinner:

PR #113, opened 2026-06-12
test(tags): cffile script-call form (named attributes + attributeCollection)

bpamiri:
## Cross-engine gap: the script-call form of `cffile` does not exist

CFML exposes built-in tags as script-callable functions. Inside a function:

```cfml
cffile(action = "read", file = path, variable = "local.content");

// or, the generic-wrapper shape:
local.args = {action: "write", file: path, output: content};
cffile(attributeCollection = "#local.args#");
```

On RustCFML the call site is parsed as a bare identifier, so every script-form `cffile(...)` throws at runtime — the same error for the named-attribute form, the attributeCollection form, and every action:

```
Variable 'cffile' is undefined
```

| engine | `cffile(action=, ...)` script call | `cffile(attributeCollection=)` | tag `<cffile>` | `fileRead()`/`fileWrite()` BIFs |
|--------|-----------------------------------|--------------------------------|----------------|--------------------------------|
| RustCFML 0.130.0 | **`Variable 'cffile' is undefined`** | **`Variable 'cffile' is undefined`** | works | work |
| Lucee 5.4.8.2 | works | works | works | work |

The tag form and the file BIFs both work, so only the script-call lowering for `cffile` is missing. And the script-call form of `cfdirectory(...)` already works (`test_cfdirectory_function_form.cfm` passes), so this is a per-tag gap, not a general script-call gap.

### What the test asserts

- (1) `cffile(action="write", file=, output=)` creates the file with the given content
- (2) `cffile(action="read", file=, variable="local.content")` delivers the content into the calling function's local scope
- (3) `cffile(action="append", ..., addnewline="no")` appends to an existing file
- (4) `cffile(attributeCollection = "#local.args#")` executes the collected attributes (write shape)
- CONTROLS (pass on both engines): a `fileWrite()`/`fileRead()` BIF round-trip, and a tag-island `<cffile action="write">` + `<cffile action="read">` round-trip (mixed tag/script in one test file — the `test_tags_include.cfm` pattern). The tag control passing while (1)-(4) fail pins the gap to the script-call surface, not the tag's file machinery.

All scratch files live under `getTempDirectory()` and are deleted on both the pass and the fail path — self-contained and runner-safe.

### Why it matters for Wheels

Wheels' `Global.cfc` `$file()` is byte-for-byte assertion (4):

```cfml
public any function $file() {
	local.args = {};
	for (local.key in arguments) {
		local.args[local.key] = arguments[local.key];
	}
	cffile(attributeCollection = "#local.args#");
}
```

Its callers are the next rungs of the ladder: the migrator writes generated migration files (`migrator/AutoMigrator.cfc`, `action="write"`) and emits SQL files (`migrator/Base.cfc`, `action="append"` with `addNewLine`), and `renderWith(writeToFile=...)` persists rendered output (`controller/miscellaneous.cfc`, `action="write"`). None of these sit on the CRUD path that now serves, which is why this gap lived behind a local workaround since before the test-PR strategy existed — but the migrator ladder hits it immediately.

### Verification

- **RustCFML 0.130.0** (release binary), staged repo layout (`harness.cfm` + this test): **2/6** — assertions (1)-(4) all fail with `Variable 'cffile' is undefined`; both controls PASS. Identical 2/6 with `RUSTCFML_JIT=0` and with `RUSTCFML_JIT_THRESHOLD=1` — interpreter semantics, not JIT-specific.
- **Lucee 5.4.8.2** (CommandBox): all **6/6** PASS. (Adobe CF was not run for this PR; the script-call tag surface is long-standing shared CFML behavior.)
- Full `tests/runner.cfm` with this test registered (RustCFML 0.130.0): **3836/3840 across 451 suites** — the only 4 failures are this suite's gap assertions; no abort. "Variable 'cffile' is undefined" is a catchable runtime error, so the runner's try/include isolation holds.
- Runtime gap → registered in the tags block of `tests/runner.cfm`, directly after `test_cfdirectory_function_form.cfm`. Full `tests/runner.cfm` on this branch (RustCFML 0.130.0): **3836/3840 across 451 suites** — the only 4 failures are this suite's gap assertions; no abort.

### DIFF
--- a/tests/runner.cfm
+++ b/tests/runner.cfm
@@ -218,6 +218,8 @@ try { include "tags/test_cfquery_result_delivery.cfm"; } catch (any e) { writeOu
 try { include "tags/test_cfdbinfo.cfm"; } catch (any e) { writeOutput("ERROR | tags/test_cfdbinfo.cfm | " & e.message & chr(10)); }
 try { include "tags/test_cfdirectory_mapping_path.cfm"; } catch (any e) { writeOutput("ERROR | tags/test_cfdirectory_mapping_path.cfm | " & e.message & chr(10)); }
 try { include "tags/test_cfdirectory_function_form.cfm"; } catch (any e) { writeOutput("ERROR | tags/test_cfdirectory_function_form.cfm | " & e.message & chr(10)); }
+// cffile script-call form (named attributes + attributeCollection)
+try { include "tags/test_cffile_script_form.cfm"; } catch (any e) { writeOutput("ERROR | tags/test_cffile_script_form.cfm | " & e.message & chr(10)); }
 try { include "tags/test_cfdirectory_recurse_symlink.cfm"; } catch (any e) { writeOutput("ERROR | tags/test_cfdirectory_recurse_symlink.cfm | " & e.message & chr(10)); }
 try { include "tags/test_cfhttpparam_runtime_body.cfm"; } catch (any e) { writeOutput("ERROR | tags/test_cfhttpparam_runtime_body.cfm | " & e.message & chr(10)); }
 try { include "tags/test_cfmail_runtime_body.cfm"; } catch (any e) { writeOutput("ERROR | tags/test_cfmail_runtime_body.cfm | " & e.message & chr(10)); }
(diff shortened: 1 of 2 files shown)


alexskinner:
manual fixed

PR #114, opened 2026-06-13
Add CFML test for per-application session namespacing

Blute:
## What

Pins that session identity is the composite **(application name, session id)** — Lucee parity. `Application.cfc`'s pseudo-constructor runs on every request and `this.name` selects the application, so one server instance can host several applications (e.g. resolved from the Host header: hub.domain.com, delivery.domain.com, floor.domain.com) with isolated application AND session scopes. The same CFID presented under two different `this.name` values must resolve to two independent sessions, and `onSessionStart` must fire per (application, session).

## Why

RustCFML already gets the application side right — `ApplicationStore` is keyed by application name. Sessions, however, are keyed by **bare session id**: `SessionStore.get/set/remove/rotate` take only the id (`crates/cfml-vm/src/session_store.rs`), and the request handler resolves the single global `CFID` cookie with no application context (`crates/cli/src/lib.rs`, cookie extraction ~1342 / Set-Cookie ~1413).

**Consequence:** two applications in one process that receive the same CFID share — and corrupt — one session. Not hypothetical: browsers don't scope cookies by port, so localhost:8180 (app A) and localhost:8181 (app B) send the identical CFID; any copied CFID is also valid across applications (a session-fixation-shaped hole). Verified on current main:

```
app A write (no cookie)        -> started=[A] x=[alpha]  CFID minted
app B write (same CFID)        -> started=[A] x=[beta]   <- inherited A's session, no onSessionStart for B
app A read  (same CFID)        -> x=[beta]               <- B's write clobbered A
```

Notably, the new datasource session store's table is already keyed `(cfid, app_name)` — but `app_name` is a fixed per-store value ("v1 uses a single partition per store", `crates/cli/src/session/datasource.rs`), because the `SessionStore` API has no way to receive the per-request application name. The storage schema anticipates exactly this fix.

## Coverage

Two fixture apps with distinct `this.name` under `tests/lifecycle/session_namespace/{a,b}`; behavioral round-trip (runs served, self-skips from the CLI). Every probe **writes** to the session, so the test is agnostic to the v0.131.0 lazy default (record/cookie/onSessionStart on first write).

- **controls** (pass today): app A mints a cookie on first write; `onSessionStart` ran in A; A's session value persists in-app.
- **gaps** (expected-fail on current upstream): same CFID written in app B fires `onSessionStart` for (app B, id) — currently inherits app A's session; app B's write must not leak into app A; app B keeps its own value independently.

Not covered (deliberately): timing-based expiry isolation — per-app `this.sessionTimeout` on the same CFID — flaky in CI; the composite key makes it structural anyway.

Query params are passed via `cfhttpparam type="url"` because multi-param URLs currently break cfhttp — companion test PR incoming.

Introduce a typed composite key (e.g. `SessionKey { app_name, session_id }`, or an `app_name` parameter on the `SessionStore` trait methods) rather than string-concatenating at call sites — make it impossible to look up a session without an application name. Thread the current application name (already resolved for the application scope) through the handler's session resolution. All backends namespace consistently: in-memory map key, memcached key prefix (`key_prefix + app + id` — note this invalidates existing memcached session keys; worth a migration note), cluster CRDT doc keys, and the datasource store's existing `app_name` column switching from store-level constant to the real per-request value. Expiry bookkeeping (`take_expired`, the new background reaper) is then naturally per-(app, id), so per-app `this.sessionTimeout` applies to that app's sessions. Single-application deployments keep working unchanged apart from persisted key shape.

### DIFF
--- a/tests/lifecycle/session_namespace/a/Application.cfc
+++ b/tests/lifecycle/session_namespace/a/Application.cfc
@@ -1,0 +1,9 @@
+<cfcomponent output="false">
+    <cfset this.name = "lucee_compat_session_ns_a" />
+    <cfset this.sessionManagement = true />
+    <cfset this.sessionTimeout = createTimespan(0, 0, 10, 0) />
+
+    <cffunction name="onSessionStart" output="false">
+        <cfset session.started_in = "A" />
+    </cffunction>
+</cfcomponent>
--- a/tests/lifecycle/session_namespace/a/page.cfm
+++ b/tests/lifecycle/session_namespace/a/page.cfm
@@ -1,0 +1,4 @@
+<cfif (url.op ?: form.op ?: "") eq "write">
+    <cfset session.x = url.val ?: form.val ?: "" />
+</cfif>
+<cfoutput>app=[#getApplicationMetadata().name ?: ''#];started=[#session.started_in ?: ''#];x=[#session.x ?: ''#]</cfoutput>
--- a/tests/lifecycle/session_namespace/b/Application.cfc
+++ b/tests/lifecycle/session_namespace/b/Application.cfc
@@ -1,0 +1,9 @@
+<cfcomponent output="false">
+    <cfset this.name = "lucee_compat_session_ns_b" />
+    <cfset this.sessionManagement = true />
+    <cfset this.sessionTimeout = createTimespan(0, 0, 10, 0) />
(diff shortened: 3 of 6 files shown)


alexskinner:
Closed and fixed

PR #115, opened 2026-06-13
Add CFML test for literal query strings in cfhttp url

Blute:
## What

Pins that a literal query string in the cfhttp `url` attribute — `url="...page.cfm?a=1"` or `?a=1&b=2` — round-trips like any other URL (Lucee parity).

## Why

On RustCFML a cfhttp request whose url contains a literal query string fails **silently**: no exception, the result struct just comes back with empty `statuscode` and `filecontent`. Observed deterministically when cfhttp runs in a request context (the served suite — 4/4 failures every run) and intermittently from the CLI.

Diagnostics already done:
- The request line **reaches the server intact** (verified against the server debug log, query string included) — the failure is on the client's side of the exchange.
- String interpolation is not the culprit: `"#t#?a=1&b=2"` builds the correct string.
- The same parameters via `cfhttpparam type="url"` work fine (that's the test's control, and the workaround we used in #114's test).

Found while writing the session-namespacing test (#114): the fixture round-trips silently returned nothing as soon as the URL carried `?op=write&val=alpha`.

## Coverage

Local echo target + served round-trip (self-skips from the CLI), following the existing cfhttp round-trip pattern.

- **control** (passes today): two params via `cfhttpparam type="url"`.
- **gaps** (expected-fail on current upstream): single literal query param (status + value arriving), two literal query params (status + both values arriving).

Validated on current main: 4 gap assertions fail, control passes, no collateral (3851/3855).

Client side of `fn_cfhttp` — the URL with a query string is accepted and the request is emitted correctly on the wire, but the response never populates the result struct (and no error surfaces). Likely the response read/parse path keyed off the request URL; worth also surfacing whatever error is being swallowed, since the silent empty result made this expensive to trace.


alexskinner:
manual fix thanks for the commit

PR #117, opened 2026-06-13
Add session-store test for nested writes into an assigned session sub-struct

Blute:
## What

A `cfml-vm` integration test (`crates/cfml-vm/tests/session_nested_writeback.rs`) asserting that nested writes into an already-assigned `session` sub-struct are present in the `SessionData` handed to `SessionStore::set`. The CFML shape is the ubiquitous "build an auth struct incrementally" one:

```cfml
session.auth = {};                 // top-level assign
session.auth.isLoggedIn = true;    // nested write
session.auth.user        = "mat";  // nested write
```

It uses a `SpyStore` wrapping `MemoryStore` (same pattern as `session_writeback.rs`) to capture the persisted payload — no external service needed.

## Why

On current main the persisted `SessionData` carries only the empty `{}` from the first line; the nested keys are dropped:

```
test nested_session_writes_are_persisted ... FAILED
nested write session.auth.user was dropped from the persisted payload — auth persisted as []
```

Any app on a **serialising** session store (memcached or the datasource store) therefore loses the nested keys on the very next request — a framework that does `session.auth = {}` then `session.auth.isLoggedIn = true` finds the user logged out immediately. (A `MemoryStore` app masks it: the live struct survives in-process across requests, so the defect only shows in what gets *persisted*.)

## Regression window

Passed up to **v0.135.0**, broke at **v0.136.0** (`9dd2c74`, "auto-vivify nested writes on scope-qualified undeclared variables", PR #111). Bisected with this exact test:

```
v0.135.0 -> ok
v0.140.0 -> FAILED (auth persisted as [])
```

It also reproduces end-to-end: a Moopa app on the datasource store logs in (writes `session.auth.X`), and the `cf_session_data.data` blob comes back `{"auth":{}, ...}` — every protected route then 401s.

## Why a store-layer test, not a CFML-suite test

Within a single request the `session` scope is a live reference, so the nested keys read back fine (already covered by `tests/core/test_session_scope_persist.cfm`). The defect is purely in the *persisted* payload, which the memory-store CFML suite can't observe — hence an integration test that inspects the `SessionData` given to `set()`.

## Where the fix goes

The session-scope write path that marks the scope dirty / assembles the `SessionData` for persistence: a nested write (`session.auth.isLoggedIn = …`) into an already-assigned sub-struct must flag the session so the *current* `auth` value (with the nested keys) is what gets serialized — as it did pre-v0.136. The v0.136 auto-vivify change for scope-qualified nested writes appears to have detached the nested mutation from the persisted session snapshot.

### DIFF
--- a/crates/cfml-vm/tests/session_nested_writeback.rs
+++ b/crates/cfml-vm/tests/session_nested_writeback.rs
@@ -1,0 +1,206 @@
+//! Regression: nested writes into an already-assigned `session` sub-struct
+//! must be present in the `SessionData` handed to `SessionStore::set`.
+//!
+//! The CFML shape is the ubiquitous "build an auth struct incrementally" one:
+//!
+//! ```cfml
+//! session.auth = {};                 // top-level assign
+//! session.auth.isLoggedIn = true;    // nested write
+//! session.auth.user        = "mat";  // nested write
+//! ```
+//!
+//! Lucee/ACF persist the whole `auth` struct. The value committed to the
+//! store must therefore carry `isLoggedIn` and `user` — not just the empty
+//! `{}` from the first line.
+//!
+//! Why this is a store-layer test and not a CFML-suite test: within a single
+//! request the `session` scope is a live reference, so the nested keys are
+//! readable regardless (covered by tests/core/test_session_scope_persist.cfm).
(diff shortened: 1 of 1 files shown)


alexskinner:
Fix submitted closing

PR #118, opened 2026-06-13
Add CFML test for WITH (CTE) statements as row-returning queries

Blute:
## What

A self-contained SQLite test (`tests/tags/test_cte_with_query.cfm`, bundled-driver, same skip pattern as `test_queryexecute_param_structs.cfm`) asserting that a `WITH` (CTE) statement is run as a row-returning query.

## Why

CTEs are standard SQL — Lucee runs `WITH … SELECT …` as an ordinary query and returns its rows. RustCFML decides whether a statement returns rows by looking for a leading `SELECT`, so a `WITH` prefix is misclassified as a non-row "execute" statement:

- **SQLite**: errors `Execute returned results - did you mean to call query?`
- **PostgreSQL**: the row is counted but every column reads back NULL (so e.g. `deserializeJSON(q.data)` gets `""` → "EOF while parsing").

On current main (v0.143.0):

```
FAIL | queryExecute: WITH (CTE) statements are row-returning queries | 1/4 passed (3 failed)
       FAIL: inline-VALUES CTE query failed (misclassified as execute?): queryExecute: SQL error: Execute returned results - did you mean to call query?
       FAIL: table CTE query failed (misclassified as execute?): ...
       FAIL: CTE-with-aggregate query failed: ...
```

(The plain-`SELECT` control passes, confirming the harness/datasource is fine.)

This is **not** a regression — it reproduces identically on v0.132/v0.135/v0.143; CTEs simply hadn't been exercised before. Real-world hit: the Moopa sysadmin route-security panel builds its payload with `WITH profile_subjects AS (…), role_subjects AS (…) SELECT row_to_json(…)::text AS data …`; on PostgreSQL the `data` column comes back empty and `deserializeJSON("")` 500s the endpoint.

## Coverage

- control: a plain `SELECT` returns rows (passes today)
- inline-`VALUES` CTE returns its rows
- CTE over a real table returns its rows **and** column values (catches the PostgreSQL NULL-column form, not just the SQLite throw)
- CTE feeding an aggregate (`count(*)`)

## Where the fix goes

The statement classifier that decides query-vs-execute (and, on the pg path, whether to fetch + decode result rows): a statement whose first significant keyword is `WITH` should be treated as row-returning, same as `SELECT`. Practically: after stripping leading whitespace/comments, accept a leading `WITH` (a CTE that resolves to a `SELECT`) as a query — ideally also handle data-modifying CTEs (`WITH … INSERT/UPDATE … RETURNING`).

### DIFF
--- a/tests/runner.cfm
+++ b/tests/runner.cfm
@@ -224,6 +224,7 @@ try { include "tags/test_pg_vector_param_binds.cfm"; } catch (any e) { writeOutp
 try { include "tags/test_pg_error_cause_chain.cfm"; } catch (any e) { writeOutput("ERROR | tags/test_pg_error_cause_chain.cfm | " & e.message & chr(10)); }
 try { include "tags/test_pg_extended_param_binds.cfm"; } catch (any e) { writeOutput("ERROR | tags/test_pg_extended_param_binds.cfm | " & e.message & chr(10)); }
 try { include "tags/test_cfquery_quoted_identifier.cfm"; } catch (any e) { writeOutput("ERROR | tags/test_cfquery_quoted_identifier.cfm | " & e.message & chr(10)); }
+try { include "tags/test_cte_with_query.cfm"; } catch (any e) { writeOutput("ERROR | tags/test_cte_with_query.cfm | " & e.message & chr(10)); }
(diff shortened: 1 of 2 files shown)


alexskinner:
thanks for the pr, fixed and merged

PR #125, opened 2026-06-14
test(pg): pool must not validate (SELECT 1) on every checkout — Lucee parity / remote-DB perf

Blute:
## Lucee-compat / performance gap

Lucee's JDBC datasource pool does **not** run a liveness query on checkout by default (`validate` defaults off), so a `cfquery` is **one** round-trip to the server.

RustCFML's Postgres pool (`r2d2` `PostgresConnectionManager`) leaves r2d2's `test_on_check_out` at its default (**true**), and `is_valid` is implemented as:

```rust
// crates/cfml-stdlib/src/builtins.rs  (~6526)
fn is_valid(&self, conn) -> ... {
    conn.simple_query("SELECT 1").map(|_| ())...
}
```

Because a connection is checked out **per query** (`execute_postgres` → `pool.get()` per `cfquery`, ~7424), **every** `cfquery` fires an extra `SELECT 1` before the real statement → **two round-trips per query**. Against a remote DB (e.g. Neon) this roughly **doubles** request latency versus Lucee on the *same* database. (Found while comparing two Moopa apps on one Neon instance — the Lucee app was ~2× faster purely from this.)

## Test

Adds `tests/tags/test_pg_pool_checkout_validation.cfm`, gated on `RUSTCFML_TEST_PG_URL` **and** the presence of `pg_stat_statements` (the only CFML-observable signal for an internal pool ping). After a stats reset it runs one structurally-distinct marker query, then asserts **zero** bare `SELECT 1` (normalized `SELECT $1`) calls were recorded.

Run locally:
```bash
docker run --rm -d -p 5432:5432 -e POSTGRES_PASSWORD=postgres postgres:16 \
  -c shared_preload_libraries=pg_stat_statements
RUSTCFML_TEST_PG_URL=postgresql://postgres:postgres@127.0.0.1:5432/postgres \
  cargo run -- tests/runner.cfm
```

**RED on stock v0.149.0** — records **2** validation calls after a single `cfquery`:
```
FAIL | PostgreSQL pool checkout validation
   FAIL: PG pool must not ping SELECT 1 on checkout (Lucee parity);
         recorded 2 validation query call(s) after one cfquery
```

## Suggested engine fix (yours to make)

One line in `get_postgres_pool`'s builder (`crates/cfml-stdlib/src/builtins.rs` ~6719):

```rust
r2d2::Pool::builder()
    .max_size(10)
    .min_idle(Some(1))
    .connection_timeout(Duration::from_secs(30))
    .test_on_check_out(false)   // <-- don't ping on every checkout
    .build(mgr)
```

`has_broken()` already returns `false`; with validation off, a connection killed server-side surfaces on the next real statement instead of costing a round-trip on every checkout. (If you'd rather detect dead connections without the per-checkout cost, `has_broken` / an `idle`-recycle is the usual r2d2 route.)

Test-only PR — no engine changes here.

### DIFF
--- a/tests/runner.cfm
+++ b/tests/runner.cfm
@@ -232,6 +232,7 @@ try { include "tags/test_pg_jsonb_param_binds.cfm"; } catch (any e) { writeOutpu
 try { include "tags/test_pg_vector_param_binds.cfm"; } catch (any e) { writeOutput("ERROR | tags/test_pg_vector_param_binds.cfm | " & e.message & chr(10)); }
 try { include "tags/test_pg_error_cause_chain.cfm"; } catch (any e) { writeOutput("ERROR | tags/test_pg_error_cause_chain.cfm | " & e.message & chr(10)); }
 try { include "tags/test_pg_extended_param_binds.cfm"; } catch (any e) { writeOutput("ERROR | tags/test_pg_extended_param_binds.cfm | " & e.message & chr(10)); }
+try { include "tags/test_pg_pool_checkout_validation.cfm"; } catch (any e) { writeOutput("ERROR | tags/test_pg_pool_checkout_validation.cfm | " & e.message & chr(10)); }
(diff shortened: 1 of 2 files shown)


alexskinner:
fixed thanks for the pr

PR #126, opened 2026-06-14
test(tags): cfinvoke argumentCollection forwards positional (numeric-keyed) args

bpamiri:
## Gap: `cfinvoke` argumentCollection drops positional (numeric-keyed) args

`cfinvoke`'s `argumentCollection` is a full argument collection, exactly like `fn(argumentCollection=st)` and the `invoke()` BIF: NUMERIC string keys (`"1"`, `"2"`, …) forward as **positional** arguments, named keys as named arguments. Lucee, Adobe CF, and BoxLang all honor numeric keys positionally. RustCFML 0.153.0 forwards the named keys but silently **drops** the numeric/positional keys — the callee's positional param keeps its default:

```cfml
function take(any a, string named = "DEF") {
    return "a=" & (isStruct(a) ? "struct:" & a.body : toString(a)) & "|named=" & named;
}
cfinvoke(component=o, method="take", returnVariable="r",
         argumentCollection = { 1: {body:"POS"}, named: "N" });
```

| engine | `r` |
|--------|-----|
| RustCFML 0.153.0 | `"a=\|named=N"` — positional `1` dropped (param keeps default) |
| Lucee 5.4.8.2 | `"a=struct:POS\|named=N"` — positional `1` → `a` |

All-named `argumentCollection` already forwards correctly on both engines (asserted as the control).

### What the test asserts
- A numeric-keyed `argumentCollection` entry arrives as the positional arg. *(red on RustCFML)*
- The positional struct survives alongside a named key — the Wheels `$invoke` shape. *(red on RustCFML)*
- CONTROL (green on both engines): all-named `argumentCollection` forwards correctly.

### Why it matters for Wheels
`Global.cfc::$invoke()` ends in `cfinvoke(attributeCollection="#local.args#")`, where `local.args.argumentCollection` is rebuilt from a dynamic call's positional args as a **numeric-keyed** struct (Global.cfc ~289–305). This is the dispatch path behind `onMissingMethod`, so association dynamic methods lose their positional argument: `post.createComment(commentStruct)` injects the foreign key as a *named* key (survives) but passes the attributes as a *positional* struct (dropped), so `validatesPresenceOf` fails; and `dependent="delete"` cascade (`deleteAll<Assoc>()`) never receives its args. Surfaced laddering hasMany/belongsTo associations on the blog tutorial app — the blog CRUD itself works, this is the association-create / dependent-delete layer.

### Verification
- **RustCFML 0.153.0** (release binary, full `tests/runner.cfm`): 1/3 — the two positional-forwarding assertions fail with `a=(null)`; the all-named control passes. Suite contributes exactly its 2 failures, no abort (runtime gap). Full runner 3863/3865 across 468 suites.
- **Lucee 5.4.8.2** (CommandBox): 3/3 PASS.
- Registered in the tags block after `test_cfinvoke_statement.cfm`.

### DIFF
--- a/tests/runner.cfm
+++ b/tests/runner.cfm
@@ -404,6 +404,16 @@ try { include "core/test_component_soft_keyword.cfm"; } catch (any e) { writeOut
 //     also accepts the ACF-style cf-prefixed `cfinvoke` spelling, but Lucee
 //     rejects it, so the cross-engine test uses the cf-less `invoke`.)
 try { include "tags/test_cfinvoke_statement.cfm"; } catch (any e) { writeOutput("ERROR | tags/test_cfinvoke_statement.cfm | " & e.message & chr(10)); }
+//   - cfinvoke_argumentcollection_positional: cfinvoke's argumentCollection is
+//     a full argument collection — NUMERIC string keys forward as POSITIONAL
+//     args (named keys as named), like fn(argumentCollection=st) / invoke().
+//     RustCFML forwards named keys but DROPS numeric-keyed (positional)
+//     entries. Wheels' Global.cfc $invoke() rebuilds the dynamic call's
+//     positional args as a numeric-keyed argumentCollection, so onMissingMethod
+//     association methods (post.createComment(struct), dependent=delete's
+//     deleteAll<Assoc>()) lose their positional arg — the FK (named) survives,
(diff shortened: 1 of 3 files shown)


alexskinner:
thanks for the pr closing

PR #129, opened 2026-06-14
test(oop): onMissingMethod must preserve named-argument NAMES in missingMethodArguments

bpamiri:
## Gap

When a method call falls through to `onMissingMethod`, the engine hands the call's arguments to the handler as `missingMethodArguments`. The KEYS of that struct depend on how the method was called:

- **NAMED** call `obj.probe(label="world")` → keyed by NAME (`"label"`)
- **POSITIONAL** call `obj.probe("world")` → keyed by POSITION (`"1"`)

Lucee 5/6/7 and Adobe ColdFusion honor both shapes. RustCFML 0.153.0 keys the NAMED-call struct by numeric position as well, dropping the argument names entirely.

```cfm
component {
  function onMissingMethod(missingMethodName, missingMethodArguments){
    return structKeyList(missingMethodArguments);
  }
}
obj.probe(label = "world")          // Lucee: "label"      | RustCFML 0.153.0: "1"
obj.probe(label = "a", extra = "b") // Lucee: "label,extra" | RustCFML 0.153.0: "1,2"
```

A positional call (`obj.probe("world")`) is correct on both engines — that's the CONTROL.

| Call shape | Lucee 5.4.8.2 | RustCFML 0.153.0 |
|---|---|---|
| `probe(label="world")` | `structKeyList` = `label`; `mma.label` = `world` | `structKeyList` = `1`; `mma.label` undefined |
| `probe(label="a", extra="b")` | keys `label,extra` | keys `1,2` |
| `byStatus(status="published")` | `mma.status` = `published` | `mma.status` undefined |
| `probe("world")` (positional CONTROL) | key `1` = `world` | key `1` = `world` |

## What the test asserts

- Single named arg → the NAME key (`label`) is present, its value round-trips, and no numeric `1` key is substituted for the name.
- Two named args → BOTH names (`label`, `extra`) are present with correct values; `StructCount == 2`.
- A Wheels-shaped dynamic scope `byStatus(status="published")` → `missingMethodArguments.status` is readable BY NAME.
- CONTROL: a positional call still uses numeric keys (`1`) on both engines and carries no `label` name — guards the wiring so a named-arg regression can't masquerade as this gap.

Assertions key on PRESENCE/VALUE, never key order (Lucee uppercases and reorders struct keys).

## Why it matters for Wheels

Wheels' dynamic query scopes use the documented pattern:

```cfm
scope(name="byStatus", handler="scopeByStatus");
model("Post").byStatus(status="published");
```

That named call dispatches through `onMissingMethod`, and the handler reads the option BY NAME (`arguments.status`). On RustCFML the struct arrives as `{1="published"}`, so `arguments.status` is undefined and the scope's `WHERE` clause is built empty — the query returns the wrong rows (0, or all). Surfaced laddering named query scopes.

## Siblings (no assertion overlap)

This is distinct from:
- **#126** — `cfinvoke argumentCollection` positional binding
- **#95** — `invoke()` dropping undeclared argument-struct keys
- **#82** — named-arg numeric-alias leak on a DECLARED-param call

Those concern the `invoke()`/`cfinvoke` marshaling path and declared-param frames. This gap is specifically `onMissingMethod` failing to preserve the NAMES of named arguments in `missingMethodArguments`.

## Verification

- **RED** — RustCFML 0.153.0 (`/tmp/rustcfml-test/rustcfml-0.153.0`), staged with harness + run.cfm + fixture: `SUMMARY: 4/16 passed` / `FAILED: 12` on the default JIT, `RUSTCFML_JIT=0`, and `RUSTCFML_JIT_THRESHOLD=1`. The only 4 passing assertions are the positional CONTROL asserts.
- **GREEN** — Lucee 5.4.8.2 via CommandBox 6.3.2 (`box task run Task.cfc`, `assert`→`chk` to avoid CommandBox's `assert` shadow, test included inside `savecontent`, version from `server.lucee.version`): `PASS: 16  FAIL: 0  ALL GREEN`.
- Full `tests/runner.cfm` on this branch (RustCFML 0.153.0): **3907/3919 across 469 suites** — the only 12 failures are this suite's named-key assertions; no abort.

### DIFF
--- a/tests/core/OnMissingMethodNamedKeysFixture.cfc
+++ b/tests/core/OnMissingMethodNamedKeysFixture.cfc
@@ -1,0 +1,49 @@
+// Fixture for the onMissingMethod named-argument-key test.
+//
+// onMissingMethod receives the call's arguments in `missingMethodArguments`.
+// For a NAMED-argument call (obj.probe(label="world")) the engines key that
+// struct by the argument NAMES; for a POSITIONAL call (obj.probe("x")) they
+// key it by numeric position ("1", "2", ...). RustCFML 0.153.0 keys the
+// named-call struct by numeric position too, losing the names — so a handler
+// that reads missingMethodArguments.label sees nothing.
+//
+// Each probe reports the shape of missingMethodArguments back to the caller
+// without depending on key ORDER (Lucee uppercases/reorders struct keys, so
+// any order-sensitive assertion would false-fail on a conforming engine).
+component {
+
+	// The component declares NO `probe`/`byStatus`/etc. method, so every call
+	// to one of those falls through to onMissingMethod.
+	public any function onMissingMethod(
+		required string missingMethodName,
+		required struct missingMethodArguments
+	) {
+		var mma = arguments.missingMethodArguments;
+		return {
+			// Sorted key list so the caller can do presence checks via
+			// listFindNoCase without caring about engine key order.
+			keyList   = structKeyList(mma),
+			count     = structCount(mma),
+			// Direct by-NAME reads — these are what a real handler does
+			// (arguments.status, arguments.label, ...).
+			hasLabel  = structKeyExists(mma, "label"),
+			labelVal  = structKeyExists(mma, "label") ? mma.label : "(absent)",
+			hasExtra  = structKeyExists(mma, "extra"),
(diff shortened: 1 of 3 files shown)


alexskinner:
thanks for the PR, closed and fixed

PR #130, opened 2026-06-14
test(core): this-scope nested write must auto-vivify its container (residual of #111)

bpamiri:
## Gap

Assigning to a member path of an **undeclared `this`-scope key** must auto-create the container as a struct on the first nested write, exactly as for every other scope:

```cfm
component {
    function init() {
        this.paths.migrate = "X";   // this.paths never pre-seeded
    }
}
```

On Lucee/ACF/BoxLang the engine implicitly performs `this.paths = {}` and the member lands. PR #111 (merged v0.136) fixed this auto-viv contract for `variables` / `local` / `request` — and those **still pass on 0.153.0** — but the fix **missed the `this` scope**. On RustCFML 0.153.0 the nested write to an undeclared `this.X` is **lost silently**: no throw at the write, the scope key may register (`StructKeyExists(this, "X")` → true) but it is bound to a non-struct value, so `IsStruct(this.X)` → false, the written member vanishes, and reading it back yields `""`.

| Engine | `this.X.y = v` on undeclared `this.X` |
|---|---|
| Lucee 5.4.8.2 | auto-vivifies — `IsStruct(this.X)` true, `this.X.y == v` |
| Adobe CF 2018-2025 | auto-vivifies (same contract) |
| BoxLang | auto-vivifies (same contract) |
| **RustCFML 0.153.0** | **silently lost — `IsStruct(this.X)` false, member gone** |

## What the test asserts

A fixture CFC performs the writes inside methods (the real Wheels context) and returns a diagnosis string:

- **Gap 1** — `this.X.migrate = "viv-this"` on an undeclared `this.X` → expects `migrate=[viv-this]`; RustCFML returns `NOT-A-STRUCT`.
- **Gap 2** — two-level chain `this.X.a.b = "deep-this"` must vivify every level → expects `b=[deep-this]`; RustCFML returns `OUTER-NOT-A-STRUCT`.
- **Control A** — `variables.X.k` autoviv inside the same CFC method still works (pins that PR #111 holds and this is a *residual* specific to `this`, not a regression).
- **Control B** — a **pre-initialized** `this` container takes the same nested write fine (isolates the failure to the missing implicit `this.X = {}` — i.e. auto-vivification, not `this`-scope writes in general).

Both controls pass on RustCFML 0.153.0 today; both gap assertions fail.

## Why it matters for Wheels

`vendor/wheels/Migrator.cfc` `init()` sets up its path table with no prior `this.paths = {}`:

```cfm
this.paths.migrate   = expandPath(arguments.migratePath);
this.paths.sql       = ...;
this.paths.templates = ...;
```

On RustCFML every one of those writes evaporates, so `this.paths` is not a struct and carries none of the path keys. `getAvailableMigrations()` and every other migrator entry point that reads `this.paths.*` then fails — the entire migrator is unusable. This is the first place the missed-`this`-scope gap surfaces in a real Wheels workload, and it is the direct sibling of the `variables.$class.*` shape #111 fixed for controller/model initialization.

## Verification

- **RED — RustCFML 0.153.0** (`/tmp/rustcfml-test/rustcfml-0.153.0`): 2/4 passed (2 failed) — the two gap assertions fail with `NOT-A-STRUCT` / `OUTER-NOT-A-STRUCT`; both controls pass. Reproduced identically in default mode, `RUSTCFML_JIT=0`, and `RUSTCFML_JIT_THRESHOLD=1`.
- **GREEN — Lucee 5.4.8.2** (CommandBox `box task run`, `assert`→`chk` rename, savecontent capture, `server.lucee.version` cited): 4/4 passed.

Full `tests/runner.cfm` on this branch (RustCFML 0.153.0): **3905/3907 across 469 suites** — the only 2 failures are this suite's this-scope assertions; no abort.

### DIFF
--- a/tests/core/ThisScopeAutoVivFixture.cfc
+++ b/tests/core/ThisScopeAutoVivFixture.cfc
@@ -1,0 +1,81 @@
+// Fixture for the THIS-scope nested auto-vivification test. Each probe runs
+// inside a CFC method (the Wheels Migrator.cfc init() context) and reports the
+// exact state the `this` scope ended up in, so a silent-loss engine fails with
+// a diagnosis instead of a bare value mismatch.
+//
+// Mirrors ScopedAutoVivFixture.cfc (the variables/local/request residual from
+// PR #111) -- this fixture isolates the one scope #111 MISSED: `this`.
+component {
+
+	// The exact Wheels Migrator.cfc init() shape: this.paths.migrate = "..."
+	// where this.paths is never pre-seeded. Reports container state + value.
+	public string function vivThisNested() {
+		try {
+			this.zzpThisViv.migrate = "viv-this";
+		} catch (any e) {
+			return "THREW: " & e.message;
+		}
+		if (!StructKeyExists(this, "zzpThisViv")) {
+			return "NO-CONTAINER";
+		}
+		if (!IsStruct(this.zzpThisViv)) {
+			return "NOT-A-STRUCT";
+		}
+		if (!StructKeyExists(this.zzpThisViv, "migrate")) {
+			return "KEY-LOST";
+		}
+		return "migrate=[" & this.zzpThisViv.migrate & "]";
+	}
+
+	// Two-level chain (Migrator stores several path keys; a deep mixin shape):
+	// this.X.a.b must vivify EVERY missing level as a struct.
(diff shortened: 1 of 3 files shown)


alexskinner:
thanks for pr fixed manually

PR #132, opened 2026-06-14
test(oop): bare component name resolves relative to caller package

bpamiri:
## Gap

A bare (unqualified) component name in `createObject("component", "Sibling")` (and `new Sibling()`) must resolve relative to the **calling CFC's package** first. RustCFML 0.153.0 does not search the caller package and throws `Could not find the component [Sibling]`; Lucee/Adobe CF/BoxLang all resolve the bare name against the caller package.

Minimal repro -- `oop/relcomp/Maker.cfc` and `oop/relcomp/Sibling.cfc` in the **same** package:

```cfm
// oop/relcomp/Maker.cfc
function viaCreate() {
    try {
        return createObject("component", "Sibling").hi();  // bare name
    } catch (any e) {
        return "ERR:" & e.message;
    }
}
```

```cfm
// from the test docroot
m = createObject("component", "oop.relcomp.Maker").init();
m.viaCreate();   // expected "SIBLING-OK"
```

| Engine | `createObject("component","Sibling")` from Maker | `new Sibling()` from Maker | qualified `oop.relcomp.Sibling` (control) |
|---|---|---|---|
| Lucee 5.4.8.2 | resolves -> `SIBLING-OK` | resolves -> `SIBLING-OK` | resolves |
| RustCFML 0.153.0 | `Could not find the component [Sibling]` (catchable) | `Could not find the component [Sibling]` (uncatchable) | resolves |

## What the test asserts

`tests/oop/test_relative_component_resolution.cfm` instantiates `Maker` by its fully-qualified path, then asserts:

1. **Control** -- `createObject("component","oop.relcomp.Sibling")` resolved from the test (docroot) scope returns `SIBLING-OK` (passes on both engines; proves the fixtures are well-formed). The control is resolved from test scope, not from inside `Maker`, because a *dotted* path is itself resolved relative to the caller package.
2. **Gap** -- `Maker.viaCreate()` (bare `createObject("component","Sibling")` from inside Maker, which lives in `oop.relcomp`) returns `SIBLING-OK`. RED on RustCFML: `viaCreate()`'s `try/catch` catches the resolution miss and returns `ERR:Could not find the component [Sibling].`, so it is a wrong-VALUE assertion failure with no file abort -- runner-safe.

[...]

PR #133, opened 2026-06-14
test(oop): bare component name in an inherited method resolves to the defining component's package

bpamiri:
## Gap: a bare component name in an INHERITED method resolves against the subclass's package, not the defining component's

A bare (unqualified) `CreateObject("component", "Sibling")` (or `new Sibling()`) inside a method must resolve relative to the package of the component that **lexically defines** the method. When that method is **inherited** by a subclass living in a **different directory**, the bare name must still be found next to the **parent**. Lucee/Adobe CF/BoxLang do this. RustCFML 0.160.0 resolves the bare name against the runtime **subclass's** directory and fails.

```cfml
// oop/inh/InhSibling.cfc      -> hi() returns "inh-sibling-ok"
// oop/inh/InhParent.cfc:
component {
    function viaCreate() { return CreateObject("component", "InhSibling").hi(); }
}
// oop/inhsub/InhChild.cfc  (DIFFERENT dir):
component extends="oop.inh.InhParent" {}
```

| call | RustCFML 0.160.0 | Lucee 5.4.8.2 |
| `createObject("oop.inhsub.InhChild").viaCreate()` (inherited) | **`Could not find the component [InhSibling]`** | `inh-sibling-ok` |
| `createObject("oop.inh.InhParent").viaCreate()` (caller is the definer — control) | `inh-sibling-ok` | `inh-sibling-ok` |

This is the **inherited-method sibling of #132** (which fixed bare resolution when the caller *is* the defining component, co-located with the sibling). #132 keyed resolution off the caller package correctly for the direct case; this remaining case keys off the concrete subclass's directory instead of the defining (parent) component's directory.

### What the test asserts
- Inherited `viaCreate()` invoked on a subclass in another directory resolves the bare sibling against the **parent's** package. *(red on RustCFML)*
- CONTROL (green on both engines): `viaCreate()` called directly on the defining parent resolves the bare sibling (the #132 case).

The bare-`CreateObject` form is catchable (`viaCreate()` returns `ERR:...` → a wrong-value assertion, no file abort), so the test is runner-safe.

### Why it matters for Wheels
This is the last blocker for the Wheels **migrator** on RustCFML. User migrations live in `app/migrator/migrations/` and extend `wheels.migrator.Migration`; `Migration.createTable()` does a bare `CreateObject("component", "TableDefinition")` to reach its package-sibling in `vendor/wheels/migrator/`. On RustCFML the inherited `createTable()` resolves `"TableDefinition"` against the user-migration directory and throws `Could not find the component [TableDefinition]`, so the migrator cannot create a single table. The same shape recurs across `Migration.cfc` (TableDefinition ×2, ViewDefinition, ForeignKeyDefinition) and `TableDefinition.cfc` (ColumnDefinition, ForeignKeyDefinition). Surfaced running the migrator pristine on stock v0.160.0 — every other migrator blocker (#129/#130/#131/#132) is now fixed; this is the one remaining wall. (The whole compat-matrix migrator suite exercises this idiom and runs green on Lucee/Adobe/BoxLang in CI.)

### Verification
- **RustCFML 0.160.0** (release binary, full `tests/runner.cfm`): 1/2 — the inherited-method assertion fails (`ERR:Could not find the component [InhSibling]`); the direct-parent control passes. Full runner 3910/3911, no abort.
- **Lucee 5.4.8.2** (served via `box server`, Application.cfc declaring the `/lib` + `/sub` mappings — box-task contexts don't honor `this.mappings`, so the served instance is the faithful comparison): inherited-on-subclass and control both resolve `inh-sibling-ok`.
- Registered in the OOP block after `test_chained_writeback_clobber.cfm`.

### DIFF
--- a/tests/oop/inh/InhParent.cfc
+++ b/tests/oop/inh/InhParent.cfc
@@ -1,0 +1,13 @@
+// Defines a method that reaches its package-sibling by BARE name. The bare
+// name must resolve relative to THIS component's directory (oop/inh/) — the
+// directory where the CreateObject literal is lexically defined — regardless
+// of which subclass instance the method runs on.
+component {
+    public string function viaCreate() {
+        try {
+            return CreateObject("component", "InhSibling").hi();
+        } catch (any e) {
+            return "ERR:" & e.message;
+        }
+    }
+}
--- a/tests/oop/inh/InhSibling.cfc
+++ b/tests/oop/inh/InhSibling.cfc
@@ -1,0 +1,3 @@
+// Sibling co-located with InhParent (tests/oop/inh/). Reached by a bare
+// CreateObject("component","InhSibling") from InhParent's method.
+component { public string function hi() { return "inh-sibling-ok"; } }
--- a/tests/oop/inhsub/InhChild.cfc
+++ b/tests/oop/inhsub/InhChild.cfc
@@ -1,0 +1,4 @@
+// Empty subclass in a DIFFERENT directory (oop/inhsub/) than its parent
+// (oop/inh/). Inherits viaCreate(); calling it must still resolve the bare
+// "InhSibling" against the PARENT's dir, not this subclass's dir.
+component extends="oop.inh.InhParent" {}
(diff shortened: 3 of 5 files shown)


alexskinner:
close its merged thank for pr

PR #135, opened 2026-06-15
perf(pg): prepared-statement cache — 1 round-trip/query (Lucee parity)

Blute:
**Reference prototype + diagnosis — your call on the final implementation.** This is a measured perf finding with a working fix attached, not a test-only PR (a clean behavioral test isn't really possible here — see "Testing" below).

## Symptom

A Moopa app on a remote Neon DB ran ~2× slower on RustCFML than on Lucee against the **same database**. After #125 landed (no more `SELECT 1` per checkout — thanks!), it was still ~2× slower.

## Root cause: every parameterized query is two round-trips

`run_postgres_statements` calls `client.query(stmt.sql.as_str(), …)` / `client.execute(stmt.sql.as_str(), …)` with a **raw SQL string**. In the `postgres`/`tokio-postgres` crate that awaits a fresh `prepare()` (a `Parse` round-trip) *before* it binds/executes — so every `cfquery` is **2 round-trips**. The prepared statement is then thrown away, so the next identical query re-parses.

Lucee (pgjdbc) doesn't do this: it pipelines `Parse`+`Bind`+`Describe`+`Execute`+`Sync` into a single flush (it already knows param types from the JDBC setters, so it never awaits a server `Describe`), and promotes hot statements to named server-side prepared statements after `prepareThreshold`. Net: **1 round-trip/query**.

## Measurements (same app, same Neon DB, same 7-query `GET /`)

| Engine | per query | request | round-trips/query |
|---|---|---|---|
| Lucee (pgjdbc) | ~22 ms | ~170 ms | 1 |
| RustCFML v0.161 (current) | ~48 ms | ~355 ms | 2 (prepare + execute) |
| **RustCFML + this prototype** | **~23 ms** | **~200 ms** | **1** |

Verified the pool itself is healthy (cache HIT, µs checkouts, full reuse — #125 holds); the entire gap was the extra `Parse` round-trip per query.

## The prototype

Per-connection prepared-statement cache:
- `Connection` becomes `PgConn { client, stmt_cache: HashMap<String, Statement> }` (cache is connection-scoped — server-side prepared statements belong to one session).
- `pg_prepare_cached()` prepares each SQL once, then reuses the `Statement` → bind/execute is one round-trip.
- Evict + re-prepare once on `"cached plan must not change result type"` (DDL invalidation), mirroring pgjdbc.
- Transaction path uses a transaction-local cache.

86 insertions / 15 deletions, all in `crates/cfml-stdlib/src/builtins.rs`. Ran a full Moopa app (sessions, route registry, auth) against Neon on it with no regressions.

## Testing

Unlike #125, this is **hard to assert as RED/GREEN**: re-preparing the *same* SQL is invisible to `pg_stat_statements` (identical normalized text), and protocol-level prepare counts aren't exposed via SQL. The only signals are latency (flaky) or wire-level packet counts. So I've left it as a prototype + numbers rather than a test — happy to add a best-effort timing test if you'd want one.

### DIFF
--- a/crates/cfml-stdlib/src/builtins.rs
+++ b/crates/cfml-stdlib/src/builtins.rs
@@ -6493,6 +6493,47 @@ struct PostgresConnectionManager {
     url: String,
 }
 
+/// Per-connection prepared-statement cache. tokio-postgres' `query(&str)` awaits
+/// a fresh `prepare()` (a Parse round-trip) before binding, so every
+/// parameterized query costs TWO round-trips. Caching the prepared `Statement`
+/// per connection makes the steady state ONE round-trip (bind/execute against
+/// the already-parsed statement) — matching pgjdbc/Lucee.
+#[cfg(feature = "postgres_db")]
+type PgStmtCache = HashMap<String, postgres::Statement>;
+
+/// A pooled PostgreSQL connection plus its prepared-statement cache. The cache
+/// is connection-scoped (server-side prepared statements belong to one session)
+/// and lives as long as the connection sits in the r2d2 pool.
+#[cfg(feature = "postgres_db")]
+struct PgConn {
+    client: postgres::Client,
+    stmt_cache: PgStmtCache,
+}
+
+/// Prepare `sql` once per connection, then hand back the cached `Statement`.
+#[cfg(feature = "postgres_db")]
+fn pg_prepare_cached(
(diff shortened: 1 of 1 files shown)


alexskinner:
this is fixed also sorted for SQL server which way more problems

PR #137, opened 2026-06-15
test(core): SerializeJSON must not leak internal arguments-scope sentinel keys

bpamiri:
## Gap: SerializeJSON leaks internal arguments-scope sentinel keys

RustCFML tags an arguments-derived struct with internal sentinel keys (`__arguments_scope`, `__arguments_params`). `structKeyList` / `structCount` / `structKeyExists` / `for-in` all correctly **filter** them — but `SerializeJSON` does **not**, so a struct built via `structAppend(s, arguments)` serializes the sentinels into the JSON:

```cfml
function f() { var s = {}; structAppend(s, arguments); return serializeJSON(s); }
f(from = "a", template = "welcome");
```

| engine | result |
|--------|--------|
| RustCFML 0.161.0 | `{"from":"a","template":"welcome","__arguments_scope":true,"__arguments_params":[]}` |
| Lucee 5.4.8.2 | `{"template":"welcome","from":"a"}` |

### What the test asserts
- `serializeJSON` of an arguments-derived struct omits `__arguments_scope` and `__arguments_params`, and includes the real keys. *(red on RustCFML)*
- CONTROL (green on both): `structKeyList` already filters the sentinels — establishing that the filtering exists elsewhere and only `SerializeJSON` is missing it.

### Why it matters for Wheels
Controller `sendEmail()` and many framework helpers build option structs by copying the `arguments` scope (`structAppend(s, arguments)`), then serialize or spool them. On RustCFML the leaked sentinels corrupt that output. The fix is the same filtering `structKeyList`/`Count`/`Exists`/`for-in` already apply, extended to `SerializeJSON`. (Surfaced laddering the mailer.)

### Verification
- **RustCFML 0.161.0** (release binary, full `tests/runner.cfm`): 2/4 — the two sentinel-omission assertions fail (`got: [true]`); the real-keys and structKeyList-control pass.
- **Lucee 5.4.8.2** (CommandBox): 4/4 PASS.
- Registered in the core block after `test_named_args_no_numeric_alias.cfm`.

### DIFF
--- a/tests/core/test_serializejson_arguments_sentinels.cfm
+++ b/tests/core/test_serializejson_arguments_sentinels.cfm
@@ -1,0 +1,39 @@
+<cfscript>
+suiteBegin("Core: SerializeJSON must not leak internal arguments-scope sentinel keys");
+
+// Background: RustCFML tags an arguments-derived struct with internal sentinel
+// keys (__arguments_scope, __arguments_params) — and structKeyList / structCount
+// / structKeyExists / for-in all correctly FILTER them. But SerializeJSON does
+// NOT filter them, so a struct built by structAppend(s, arguments) serializes
+// the sentinels into the JSON. Lucee has no such keys.
+//
+//   function f() { var s = {}; structAppend(s, arguments); return serializeJSON(s); }
+//   f(from="a", template="welcome")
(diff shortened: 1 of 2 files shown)


alexskinner:
this is fixed thanks for the pr test

PR #139, opened 2026-06-15
test(oop): `new Comp(args)` must propagate the exception thrown by init()

bpamiri:
## Gap

The `new Component(args)` constructor sugar must run `init(args)` and, if `init()` throws, **propagate** that exception — exactly as `createObject("component","X").init(args)` does. A CFC whose `init()` rejects bad arguments can then be trusted never to hand back a half-built object.

**RustCFML 0.161.0** runs `init()` under `new X(args)` but **swallows** any exception it throws and returns the partially-initialized object. The explicit `createObject(...).init()` path propagates correctly, so the divergence is specific to the `new` sugar.

| invocation | Lucee 5.4.8.2 | RustCFML 0.161.0 |
|---|---|---|
| `createObject("component","F").init(windowSeconds=0)` | throws | throws ✅ (CONTROL) |
| `new F(windowSeconds=30)` (valid) | builds | builds ✅ (CONTROL) |
| `new F(windowSeconds=0)` (invalid) | **throws** | **no throw — object built** ❌ |

## Test

`tests/oop/test_new_constructor_init_throw.cfm` (+ `NewInitThrowFixture.cfc`), registered in `runner.cfm`. Two CONTROL assertions (green on both engines) bracket the gap assertion. On 0.161.0: **2/3 pass** — the gap assertion (`new F(0)` must throw) fails because no exception propagates.

## Why it matters

Constructor-guard validation is defeated through the `new` path. Wheels relies on it framework-wide — e.g. `wheels.middleware.RateLimiter` `init()` throws `Wheels.RateLimiter.InvalidConfiguration` on `windowSeconds<=0` / `maxRequests<0`, and middleware is registered as `new wheels.middleware.RateLimiter(maxRequests=.., windowSeconds=..)`. On RustCFML a misconfiguration silently yields a broken, half-initialized instance instead of failing fast.

### DIFF
--- a/tests/oop/NewInitThrowFixture.cfc
+++ b/tests/oop/NewInitThrowFixture.cfc
@@ -1,0 +1,17 @@
+component {
+
+    // init() guards its argument and throws on an invalid value — the classic
+    // constructor-validation pattern. A valid value returns a usable instance.
+    function init(numeric windowSeconds = 60) {
+        if (arguments.windowSeconds <= 0) {
+            throw(type = "Fixture.InvalidConfiguration", message = "windowSeconds must be > 0");
+        }
+        variables.windowSeconds = arguments.windowSeconds;
+        return this;
+    }
+
+    public numeric function getWindowSeconds() {
+        return variables.windowSeconds;
+    }
+
+}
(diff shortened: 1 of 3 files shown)


alexskinner:

PR #140, opened 2026-06-15
test(oop): getMetadata() must not propagate a parent's displayName onto a child's leaf metadata

bpamiri:
## Gap

Component attributes such as `displayName` are **not** inherited onto a subclass's own metadata struct. `getMetadata(child)` reports the child's *declared* attributes; the parent's `displayName` lives on the parent's metadata (reachable via the metadata's `extends` chain), not copied onto the leaf. Lucee, Adobe CF and BoxLang all leave `displayName` **absent** from a child that declares none of its own.

**RustCFML 0.161.0** copies the parent's `displayName` onto the child's leaf metadata.

| call | Lucee 5.4.8.2 | RustCFML 0.161.0 |
|---|---|---|
| `getMetadata(new MetaDisplayBase())` → `displayName` | `MetaDisplayBaseLabel` | `MetaDisplayBaseLabel` ✅ (CONTROL) |
| `getMetadata(new MetaDisplayChild())` → `structKeyExists(md,"displayName")` | **false (absent)** | **true (="MetaDisplayBaseLabel")** ❌ |

## Test

`tests/oop/test_getmetadata_inherited_displayname.cfm` (+ `MetaDisplayBase.cfc` / `MetaDisplayChild.cfc`), registered in `runner.cfm`. Two CONTROL assertions (parent carries its own `displayName`; child still exposes its `extends` chain) bracket the gap assertion. On 0.161.0: **2/3 pass** — the child-leaf-must-not-carry-displayName assertion fails.

## Why it matters

Surfaced while inspecting controller metadata during package/mixin integration. Framework introspection that walks `getMetadata()` to decide behavior per concrete component (Wheels inspects component metadata at startup) must observe the leaf's **own** attributes, not silently inherited ones.

### DIFF
--- a/tests/oop/MetaDisplayBase.cfc
+++ b/tests/oop/MetaDisplayBase.cfc
@@ -1,0 +1,5 @@
+component displayName="MetaDisplayBaseLabel" {
+    function init() {
+        return this;
+    }
+}
--- a/tests/oop/MetaDisplayChild.cfc
+++ b/tests/oop/MetaDisplayChild.cfc
@@ -1,0 +1,6 @@
+component extends="oop.MetaDisplayBase" {
+    // Deliberately declares NO displayName attribute of its own.
+    function init() {
+        return this;
+    }
+}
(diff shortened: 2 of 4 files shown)


alexskinner:
Thanks for the PR fixed

PR #144, opened 2026-06-15
test(core): a named exclusive lock must be reentrant within the same request/thread

bpamiri:
## Gap

A thread that already holds an exclusive named lock must be able to re-acquire the **same** named lock without blocking — named locks are reentrant per thread. Lucee/ACF/BoxLang all let the inner block run immediately.

**RustCFML 0.161.0** is **not** reentrant: the inner re-acquire of an already-held named lock blocks until the timeout elapses, then throws a lock-timeout error. The inner block never runs.

```cfml
lock name="x" timeout="1" type="exclusive" {
    lock name="x" timeout="1" type="exclusive" { /* inner */ }
}
```

| | Lucee 5.4.8.2 | RustCFML 0.161.0 |
|---|---|---|
| outer block runs | yes | yes ✅ (CONTROL) |
| inner re-acquire | runs immediately | **times out after 1s, throws `cflock timeout`** ❌ |

## Test

`tests/core/test_named_lock_reentrant.cfm`, registered in `runner.cfm` (1s timeout to bound the failure). One CONTROL assertion (outer body runs) + two gap assertions. On 0.161.0: **1/3 pass**.

## Why it matters

Same-name re-entry is a normal CFML pattern — a locked section calling a helper that locks the same name. On RustCFML it self-deadlocks until the timeout, surfacing as request hangs/errors.

### DIFF
--- a/tests/core/test_named_lock_reentrant.cfm
+++ b/tests/core/test_named_lock_reentrant.cfm
@@ -1,0 +1,41 @@
+<cfscript>
+suiteBegin("Concurrency: a named exclusive lock is reentrant within the same request/thread");
+
+// Background: a thread that already holds an exclusive named lock must be able to
+// re-acquire the SAME named lock without blocking — named locks are reentrant per
+// thread. Lucee/ACF/BoxLang all let the inner block run immediately.
+//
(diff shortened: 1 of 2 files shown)


alexskinner:
thanks for the pr, fixed and closing

PR #150, opened 2026-06-16
RED test: cfexit method="exittag" must stop custom tag execution

Blute:
## What

Adds a CFML regression test for `<cfexit method="exittag">` inside the start phase of a body custom tag.

The fixture records that the custom tag start phase began, calls `cfexit method="exittag"`, then has code after the `cfexit` and an end-phase branch that should not run. The caller wraps the custom tag in `cfsavecontent` and asserts:

- `cfexit method="exittag"` does not throw
- the caller body content is not emitted
- code after `cfexit` and the custom tag end phase are skipped

## Why

Lucee supports this shape and Moopa uses it in custom tags that intentionally exit early, for example one-time include/dedupe tags. On RustCFML v0.182.0 the tag parser rejects `<cfexit>` outright, which means otherwise-valid Lucee/Moopa code either 500s or needs engine-specific shims.

This is one of the compatibility gaps hit while booting the `hello` Moopa app on RustCFML: we do not want the app/framework to carry separate Lucee-vs-RustCFML code paths for valid CFML behavior.

## Evidence

```
FAIL | Tags: cfexit | 1/3 passed (2 failed)
       FAIL: cfexit method="exittag" does not throw | expected: [] | got: [Tag <cfexit> is not implemented.]
       FAIL: start-phase cfexit method="exittag" skips remaining tag code and end phase | expected: [[start]] | got: []
SUMMARY: 3955/3957 passed across 485 suites
```

```
PASS | Tags: cfexit | 3/3 passed
```

(The full RustCFML runner is expectedly RED only for this new suite. The full Lucee runner has unrelated RustCFML-harness-specific failures, so the relevant signal is the targeted `Tags: cfexit` suite.)

## Where the fix likely goes

This probably needs both parser and custom-tag control-flow support, not only a builtin stub. The tag preprocessor currently emits body custom tags as:

```rust
__cfcustomtag_start(path, attrs);
body_script
__cfcustomtag_end();
```

For start-phase `cfexit method="exittag"`, RustCFML needs to stop the current custom tag and prevent the caller body / end phase from running. That likely means adding `cfexit` parsing plus a VM sentinel/state that `__cfcustomtag_start` can surface to the generated wrapper so the body/end calls are skipped.

This PR is test-only so the desired Lucee behavior is pinned before choosing that implementation shape.

### DIFF
--- a/tests/runner.cfm
+++ b/tests/runner.cfm
@@ -252,6 +252,7 @@ try { include "tags/test_tags_misc.cfm"; } catch (any e) { writeOutput("ERROR | 
 try { include "tags/test_tags_customtag.cfm"; } catch (any e) { writeOutput("ERROR | tags/test_tags_customtag.cfm | " & e.message & chr(10)); }
 try { include "tags/test_custom_tag_attribute_collection.cfm"; } catch (any e) { writeOutput("ERROR | tags/test_custom_tag_attribute_collection.cfm | " & e.message & chr(10)); }
 try { include "tags/test_tags_customtag_lifecycle.cfm"; } catch (any e) { writeOutput("ERROR | tags/test_tags_customtag_lifecycle.cfm | " & e.message & chr(10)); }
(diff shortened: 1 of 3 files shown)


alexskinner:

PR #151, opened 2026-06-16
RED test: cfhtmlhead/cfhtmlbody tags must parse and execute

Blute:
## What

Adds a CFML regression test for `<cfhtmlhead>` and `<cfhtmlbody>`. The test runs both tags from a custom-tag fixture and asserts they do not throw and that execution continues after the tags.

## Why

Lucee supports these tags for injecting content into the response `<head>` and body. Moopa uses them in its `cf_once` tag so route-local scripts/styles can be rendered once while still landing in the correct document region.

While booting the `hello` Moopa app on RustCFML, these tags were missing and had to be worked around locally by replacing `cfhtmlhead`/`cfhtmlbody` with a request queue plus explicit layout flush tags. That is exactly the kind of Lucee-vs-RustCFML fork we want to avoid in the app/framework.

## Evidence

```
FAIL | Tags: cfhtmlhead / cfhtmlbody | 0/2 passed (2 failed)
       FAIL: cfhtmlhead/cfhtmlbody do not throw | expected: [] | got: [Tag <cfhtmlhead> is not implemented.]
       FAIL: cfhtmlhead/cfhtmlbody continue executing the current template | expected: [ran] | got: []
SUMMARY: 3954/3956 passed across 485 suites
```

```
PASS | Tags: cfhtmlhead / cfhtmlbody | 2/2 passed
```

(The full Lucee runner has unrelated RustCFML-harness-specific failures; this PR only relies on the targeted suite result.)

## Scope

This PR intentionally pins the first compatibility contract: the tags are accepted and template execution continues. Lucee injects the actual head/body output into the HTTP response stream outside local `cfsavecontent`, so deeper placement assertions likely belong in a follow-up serve-mode/HTTP test once the basic tags exist.

## Where the fix likely goes

The tag parser currently rejects `<cfhtmlhead>` before execution. A fix likely needs parser arms for `cfhtmlhead` / `cfhtmlbody` and VM/request-output plumbing for queued head/body injection in serve mode.

This PR is test-only so the basic Lucee-compatible behavior is pinned first.

### DIFF
--- a/tests/runner.cfm
+++ b/tests/runner.cfm
@@ -252,6 +252,7 @@ try { include "tags/test_tags_misc.cfm"; } catch (any e) { writeOutput("ERROR | 
 try { include "tags/test_tags_customtag.cfm"; } catch (any e) { writeOutput("ERROR | tags/test_tags_customtag.cfm | " & e.message & chr(10)); }
 try { include "tags/test_custom_tag_attribute_collection.cfm"; } catch (any e) { writeOutput("ERROR | tags/test_custom_tag_attribute_collection.cfm | " & e.message & chr(10)); }
 try { include "tags/test_tags_customtag_lifecycle.cfm"; } catch (any e) { writeOutput("ERROR | tags/test_tags_customtag_lifecycle.cfm | " & e.message & chr(10)); }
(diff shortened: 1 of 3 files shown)


alexskinner:

PR #152, opened 2026-06-16
RED test: cfsleep tag must parse and execute

Blute:
## What

Adds a CFML regression test for `<cfsleep>`. The test runs `cfsleep time="#attributes.time#"` from a custom-tag fixture and asserts it does not throw and execution continues afterward.

## Why

Lucee supports `<cfsleep>` with a millisecond `time` attribute. Moopa uses it during `/init` to briefly let in-flight requests drain before `applicationStop()` in production.

While booting the `hello` Moopa app on RustCFML, `<cfsleep>` was missing and the framework delay had to be removed locally. That should not become an app/framework fork for valid Lucee CFML.

## Evidence

```
FAIL | Tags: cfsleep | 0/2 passed (2 failed)
       FAIL: cfsleep does not throw | expected: [] | got: [Tag <cfsleep> is not implemented.]
       FAIL: cfsleep continues executing the current template | expected: [slept] | got: []
SUMMARY: 3954/3956 passed across 485 suites
```

```
PASS | Tags: cfsleep | 2/2 passed
```

(The full Lucee runner has unrelated RustCFML-harness-specific failures; this PR only relies on the targeted suite result.)

## Where the fix likely goes

The tag parser currently rejects `<cfsleep>` before execution. A fix likely needs a parser arm that evaluates the `time` attribute at runtime and a VM/builtin implementation that sleeps for the requested milliseconds.

This PR is test-only so the basic Lucee-compatible behavior is pinned first.


alexskinner:

PR #153, opened 2026-06-16
RED test: cfset expression bodies ignore CFML comments

Blute:
## What

Adds a CFML runner test for CFML comments inside a tag-mode `<cfset>` expression body.

## Why

Lucee ignores `<!--- ... --->` comments embedded inside multi-line expression literals. RustCFML currently leaves the comment marker in the generated expression, so array/struct literals fail to parse. This surfaced while loading a Moopa table definition with section comments inside an array literal.

## Validation

- RustCFML v0.189.0 red test:
  - `Parse error ... Expected RBracket, found Bang`


alexskinner:

PR #154, opened 2026-06-16
RED test: cfloop list literal attributes stay strings

Blute:
## What

Adds a CFML runner test for tag-mode `<cfloop list="literal" index="...">`.

## Why

Lucee treats the `list` attribute as a literal string. RustCFML currently converts the literal directly into `listToArray(...)` without quoting it, so values such as date masks (`yyyy-MM-dd HH:mm:ss`) become identifiers/operators and fail to parse. This surfaced in a Moopa REAXML importer date-pattern loop.

## Validation

- RustCFML v0.189.0 red test:
  - `Parse error ... Expected RParen, found Identifier("HH")`


alexskinner:

PR #155, opened 2026-06-16
RED test: imp is a soft keyword in expression contexts

Blute:
## What

Adds a CFML runner test proving `imp` is usable as an ordinary local variable name.

## Why

Lucee treats `imp` as an identifier in expression contexts. RustCFML currently tokenizes it as `ImpKeyword`, so code like `var imp = application.lib.auth.getImpersonatorContext()` fails to parse. This surfaced while booting Moopa routes that keep impersonation context in a short `imp` local.

## Validation

- RustCFML v0.189.0 red test:
  - `Parse error ... Expected identifier (found ImpKeyword)`


alexskinner:

PR #156, opened 2026-06-16
RED test: escaped hash tag attributes stay literal

Blute:
## What

Adds a CFML runner test for escaped literal hashes in tag attributes.

## Why

Lucee preserves `##` as a literal `#` inside tag attributes. RustCFML currently converts the tag to script with a single `#`, then the CFScript parser treats it as the start of interpolation and consumes the rest of the generated script. This surfaced in a `cflog text="... ##278"` attribute during Moopa startup.

## Validation

- RustCFML v0.189.0 red test:
  - `Parse error ... Expected RParen, found Eof`


alexskinner:

PR #164, opened 2026-06-17
RED test: Application.cfc lifecycle overrides are case-insensitive

Blute:
## What

Adds a serve-mode lifecycle fixture proving an `Application.cfc` child method whose casing differs from the parent lifecycle method still overrides it.

The fixture defines the parent method as `OnApplicationStart`, the child method as `onApplicationStart`, calls `super.onApplicationStart()`, and asserts both parent and child startup state are present.

## Why

CFML method dispatch is case-insensitive. RustCFML v0.193.0 currently runs the parent lifecycle method but misses the child override when the method casing differs.

This matters for applications that extend a framework-level `Application.cfc` and use different casing for lifecycle methods across parent and child components. Under Lucee, both methods run. Under RustCFML v0.193.0, only the parent startup runs, so child-level application initialization can be skipped.

A concrete symptom is app-specific startup values not being initialized, which can lead downstream code to build incomplete URLs such as:

```text
redirect_uri=/login/callback
```

Lucee treats the lifecycle override case-insensitively and runs both methods, producing the expected child-level application state.

## Validation

- RustCFML v0.193.0 serve-mode targeted fixture:
  - `parent=parent-ran|child=missing`
- RustCFML v0.193.0 runner with the new test:
  - `FAIL | Lifecycle: Application.cfc method overrides are case-insensitive | 1/2 passed (1 failed)`
  - `child lifecycle override runs despite case mismatch | expected: [parent=parent-ran|child=child-ran] | got: [parent=parent-ran|child=missing]`
- Lucee 7.0.2.106 direct request to the same fixture:
  - `parent=parent-ran|child=child-ran`

This PR is test-only so the Lucee-compatible lifecycle dispatch behavior is pinned before implementation.

### DIFF
--- a/tests/lifecycle/application_lifecycle_case_override/Application.cfc
+++ b/tests/lifecycle/application_lifecycle_case_override/Application.cfc
@@ -1,0 +1,9 @@
+<cfcomponent extends="BaseApplication" output="false">
+    <cfset this.name = "rustcfml_lifecycle_case_override" />
+
+    <cffunction name="onApplicationStart" returntype="boolean" output="false">
+        <cfset super.onApplicationStart() />
+        <cfset application.lifecycle_case_child = "child-ran" />
+        <cfreturn true />
+    </cffunction>
+</cfcomponent>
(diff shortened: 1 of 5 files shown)


alexskinner:
closing this is merged/fixed thanks for the PR

PR #165, opened 2026-06-17
RED test: cfhttp must honor attributeCollection

Blute:
## What

Adds a server-mode CFML runner test proving `<cfhttp attributeCollection="#attrs#" ...>` reads `url`, `method`, and `timeout` from the supplied struct.

The test calls an existing local test target and asserts the request returns `200 OK` with the expected body.

## Why

Lucee supports passing `cfhttp` attributes through `attributeCollection`. RustCFML v0.193.0 currently ignores the collected `url` for tag-mode `cfhttp`, then throws:

```text
cfhttp: url is required
```

This matters for code that builds HTTP request attributes dynamically, then passes them into `cfhttp` as one struct. A representative OAuth token exchange builds a struct containing the token endpoint URL and method, then calls:

```cfml
<cfhttp attributeCollection="#stAttr#" result="stResponse">
```

Under RustCFML, the request fails before it is made because `cfhttp` does not read `stAttr.url`.

## Validation

- RustCFML v0.193.0 runner with the new test:
  - `FAIL | Tags: cfhttp attributeCollection | 0/3 passed (3 failed)`
  - `cfhttp attributeCollection url/method error | expected: [] | got: [cfhttp: url is required]`
- Lucee 7.0.2.106 direct compatibility check against a tiny HTTP target:
  - `status=200 OK|body=echo-ok`

This PR is test-only so the Lucee-compatible `cfhttp attributeCollection` behavior is pinned before implementation.


alexskinner:
this is manually fixed thanks for the PR

PR #167, opened 2026-06-17
RED test: SessionCommit must be callable

Blute:
## What

Adds a core test proving `SessionCommit()` is callable and leaves existing session data intact.

## Why

Lucee exposes `SessionCommit()` as a built-in function. RustCFML v0.200.0 currently treats it as an undefined variable/function:

```text
Variable 'SessionCommit' is undefined
```

This matters for request flows that mutate `session` and then immediately redirect or abort. `SessionCommit()` is used to force the current session state to persistent storage before that control-flow boundary.

## Validation

- RustCFML v0.200.0 targeted suite:
  - `FAIL | SessionCommit | 1/2 passed (1 failed)`
  - `SessionCommit is callable | expected: [] | got: [Variable 'SessionCommit' is undefined]`
- Lucee 7.0.2.106 direct check:
  - `SessionCommit OK|probe=before`

This PR is test-only so the Lucee-compatible session function behavior is pinned before implementation.

### DIFF
--- a/tests/core/test_session_commit.cfm
+++ b/tests/core/test_session_commit.cfm
@@ -1,0 +1,17 @@
+<cfscript>
+suiteBegin("SessionCommit");
+
+session.sessionCommitProbe = "before";
+sessionCommitError = "";
+
+try {
+    SessionCommit();
+} catch (any e) {
+    sessionCommitError = e.message;
+}
+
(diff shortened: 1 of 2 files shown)


alexskinner:
closed and fixed thanks for the PR

PR #168, opened 2026-06-18
fix: bind this for case-insensitive super calls

Blute:
## What

Fixes `super.method()` dispatch when the parent method casing differs from the call-site casing.

The `__super` dispatch path now resolves the parent method case-insensitively before invoking the stored parent function. The PR also adds a serve-mode `Application.cfc` lifecycle fixture where:

- the parent defines `OnApplicationStart`
- the child defines `onApplicationStart`
- the child calls `super.onApplicationStart()`
- the parent lifecycle method reads `this.parentMarker`

## Why

CFML method dispatch is case-insensitive, including `super` calls. Before this fix, RustCFML's special `__super` path used an exact-case lookup first. If that lookup missed, execution fell through to generic struct method dispatch, which can call the parent function without binding the child component as `this`.

The visible symptom is a parent lifecycle method reached through `super` failing with:

```text
Variable 'this' is undefined
```

Lucee binds `this` for the same shape, so both parent and child lifecycle startup complete.

## Validation

- Pre-fix RustCFML direct repro:
  - `ERR:Variable 'this' is undefined`
- Lucee 7 direct request to the same fixture:
  - `parent=parent-this|child=child-ran`
- RustCFML direct repro after fix:
  - `parent=parent-this|child=child-ran`
- `cargo build -p rustcfml-cli`
- RustCFML serve-mode runner after fix:
  - `SUMMARY: 4113/4113 passed across 503 suites`
- Linux aarch64 release build in Docker:
  - `RustCFML v0.203.0`

### DIFF
--- a/crates/cfml-vm/src/lib.rs
+++ b/crates/cfml-vm/src/lib.rs
@@ -5523,7 +5523,7 @@ impl CfmlVirtualMachine {
                         } else if let CfmlValue::Struct(ref s) = object {
                             if s.contains_key("__is_super") {
                                 // Super dispatch — find the parent's function by stored index
-                                let prop = object.get(&method_name).unwrap_or(CfmlValue::Null);
+                                let prop = s.get_ci(method_name.as_str()).unwrap_or(CfmlValue::Null);
                                 if let CfmlValue::Function(ref f) = &prop {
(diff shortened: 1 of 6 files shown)


alexskinner:

PR #174, opened 2026-06-19
[codex] Handle PostgreSQL DML RETURNING rows

Blute:
## Summary
- Route PostgreSQL DML statements with a top-level `RETURNING` clause through the row-returning query path.
- Preserve ordinary PostgreSQL mutation splitting for non-returning statements.
- Add a live PostgreSQL regression registered through the `cf_runtest`/`cfmodule` isolation wrapper, covering both `queryExecute(..., { returntype = "array" })` and `<cfquery returntype="array">`.

## Root Cause
RustCFML classified PostgreSQL statements as row-returning only from the first SQL keyword. `UPDATE ... RETURNING`, `INSERT ... RETURNING`, and `DELETE ... RETURNING` therefore went through `postgres::Client::execute`, which rejects statements that produce rows with `Execute returned results - did you mean to call query?`. Lucee returns those rows, and applications use this shape for atomic consume/claim operations.

## Validation
- `cargo check --features all-databases`
- Focused live regression against disposable Postgres 16 via `cfmodule template="tests/runtest.cfm" file="tags/test_pg_dml_returning.cfm"`: `16/16 passed`

## Notes
The app database user I first tried could connect but did not have `CREATE TABLE` permission in `public`, so the live validation used a disposable local Postgres container instead.

### DIFF
--- a/crates/cfml-stdlib/src/builtins.rs
+++ b/crates/cfml-stdlib/src/builtins.rs
@@ -7476,6 +7476,135 @@ fn is_select_query(sql: &str) -> bool {
     )
 }
 
+/// PostgreSQL also returns rows from DML statements with a top-level
+/// `RETURNING` clause. Those must run through `query`, not `execute`; the
+/// postgres crate rejects row-returning statements on the execute path.
+#[cfg(feature = "postgres_db")]
+fn postgres_returns_rows(sql: &str) -> bool {
+    let trimmed = strip_leading_sql_noise(sql);
+    let kw_len = trimmed
+        .as_bytes()
+        .iter()
+        .take_while(|b| b.is_ascii_alphabetic())
+        .count();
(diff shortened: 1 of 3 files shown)


alexskinner:
Landed on `main` via cherry-pick as caf96be (preserving authorship) — the GitHub merge conflicted because the branch predated main's `PgStmtCache`/`PgRunError` signature change to `run_postgres_statements`; resolved to keep that signature with this PR's `postgres_returns_rows` logic.

Followed up in v0.218.0 (ab47ff7) with the same fix for the other row-returning DML shapes: SQL Server `OUTPUT` (tiberius `execute()` silently discards the rowset) and MariaDB 10.5+ `INSERT/DELETE ... RETURNING` (`exec_drop` discards), reusing this PR's top-level keyword scanner. Thanks @Blute!

PR #175, opened 2026-06-19
Retry stale Postgres pooled connections

Blute:
## Summary
- mark pooled Postgres connections broken when query execution sees a closed driver connection
- retry once for retry-safe statements after returning the broken connection, preserving the no-SELECT-1-on-checkout behavior
- add a `cf_runtest` regression that terminates an idle PostgreSQL backend and verifies transparent retry

## Root cause
RustCFML intentionally disables r2d2 checkout validation for Lucee parity and remote database performance. That keeps each `cfquery` to one database round trip, but a server-closed idle session can still be handed back out by the pool. Neon scale-to-zero / suspend-resume closes those old sessions, so the first real `prepare` can fail with `connection closed` instead of evicting the connection and retrying.

The retry only applies when the driver reports a closed connection and the statement is retry-safe. SELECTs can retry. Mutation prepare failures can retry only before any mutation has executed. Mutation execute failures and transaction queries are not retried.

## Tests
- `cargo check -p rustcfml-cli`
- `RUSTCFML_TEST_PG_URL=postgresql://postgres:postgres@127.0.0.1:55432/postgres cargo run -- tests/runner.cfm | rg "FAIL \||FAILED:|All failures|PostgreSQL stale pooled|PostgreSQL pool checkout|SUMMARY|ALL TESTS PASSED"`
  - `PASS | PostgreSQL pool checkout validation ... | 1/1 passed`
  - `PASS | PostgreSQL stale pooled connection retry ... | 2/2 passed`
  - `SUMMARY: 4222/4222 passed across 517 suites`
  - `ALL TESTS PASSED`

## Lucee compatibility
The regression is env-gated on `RUSTCFML_TEST_PG_URL`; without a live PostgreSQL URL it skips cleanly. The runtime behavior matches Lucee-style recovery from dead idle pooled connections without adding a validation query to every checkout.

### DIFF
--- a/crates/cfml-stdlib/src/builtins.rs
+++ b/crates/cfml-stdlib/src/builtins.rs
@@ -6711,6 +6711,7 @@ type PgStmtCache = HashMap<String, postgres::Statement>;
 struct PgConn {
     client: postgres::Client,
     stmt_cache: PgStmtCache,
+    broken: bool,
 }
 
 /// Prepare `sql` once per connection, then hand back the cached `Statement`.
@@ -6768,7 +6769,7 @@ impl r2d2::ManageConnection for PostgresConnectionManager {
 
     fn connect(&self) -> Result<Self::Connection, Self::Error> {
         let client = connect_postgres(&self.url)?;
-        Ok(PgConn { client, stmt_cache: HashMap::new() })
+        Ok(PgConn { client, stmt_cache: HashMap::new(), broken: false })
     }
 
     fn is_valid(&self, conn: &mut Self::Connection) -> Result<(), Self::Error> {
@@ -6782,7 +6783,7 @@ impl r2d2::ManageConnection for PostgresConnectionManager {
         // true. `is_closed()` flips once the background connection task has
         // terminated (e.g. the server closed an idle socket), so a closed
         // connection is evicted rather than handed back out.
-        conn.client.is_closed()
+        conn.broken || conn.client.is_closed()
     }
 }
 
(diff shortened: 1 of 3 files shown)


alexskinner:
Thanks for this — the diagnosis was spot-on and the fix is now merged. While
  extending the same idea to SQL Server, live testing against real Postgres and
  SQL Server containers surfaced two things worth sharing, so I've built on your
  work in a follow-up (shipping as v0.215.0). Here's the full picture.

  What your PR got right

  The root cause is exactly as you described: with test_on_check_out(false) (which
  we keep for Lucee parity and to avoid a per-checkout round-trip), a backend
  that the server closes while idle still sits in the r2d2 pool and gets handed
  back out. has_broken only evicts it on return, so the first query to draw it
  fails. Your one-shot retry closes that window for the common case.

  What live testing revealed

  1. A single retry isn't always enough. When I tested against a real server by
  killing the connection that was just used, the retry still failed on the second
  round. Two r2d2 behaviours combine here:
  - get() does not validate on checkout (that's the whole point of
  test_on_check_out(false)), so a dead idle connection is handed out without a
  ping.
  - r2d2 establishes replacement connections asynchronously on a background
  thread.

  So when a server drops several idle sessions at once — exactly the Neon
  scale-to-zero / failover case — the pool holds multiple dead connections, and an
  immediate retry can draw a second stale one before a fresh replacement is
  ready. A single retry only survives because min_idle usually keeps one spare
  healthy connection around.

  Fix: execute_postgres (and the new execute_mssql) now retry in a bounded loop —
  discarding each broken connection so r2d2 evicts it, and trying the next pooled
  connection until a live one runs the statement. The loop is capped at the pool
  size (PG_POOL_MAX_SIZE / MSSQL_POOL_MAX_SIZE), so a statement that genuinely
  keeps failing connection-level can't spin forever.

  2. is_closed() alone misses server-side FATALs. This is the subtle one.
  is_closed() only flips once the socket is observed closed — which is what
  happens when Neon's proxy drops an idle TCP connection (your scenario, the
  connection closed at prepare). But a pg_terminate_backend, a smart/fast server
  shutdown, or a failover instead sends a server FATAL message (terminating
  connection due to administrator command, SqlState 57P01) while is_closed() is
  still false. That path wasn't recovered by either the original retry or the
  loop.

  Fix: pg_error_is_connection_fatal() now also treats operator-intervention
  (57P01/57P02/57P03) and connection-exception (class 08) SqlStates as
  connection-fatal, in addition to is_closed(). Ordinary SQL errors (syntax,
  constraint violations, etc.) are deliberately not treated as fatal, so genuine
  query failures still surface immediately rather than being retried.

  3. The regression tests were false-passing. Your env-gated test (and my first
  SQL Server one) terminated one backend and queried once — but min_idle's spare
  healthy connection meant the query never actually hit the dead connection, so
  the test passed with and without the fix. I rewrote both to kill the connection
  they just used across multiple rounds, so by round two the pool is forced to
  drain a genuinely dead connection. Verified against PostgreSQL 16 and SQL Server
  2022: each test now fails on the pre-fix code and passes with the fix.

  SQL Server

  SQL Server shared the identical defect — same r2d2 pool, same
  test_on_check_out(false), no retry — so it got the same treatment:
  run_mssql_statement now returns a rich error (connection_broken + retry_safe)
  mirroring Postgres's PgRunError, and execute_mssql runs the same bounded
  drain-loop. SELECTs always replay; mutations are never retried (tiberius fuses
  prepare+execute into a single sp_executesql round-trip, so a failure can't be
  proven to predate a write — retrying could double-apply). Transaction paths mark
  the connection broken but never retry inside a transaction.

  Net result

  - One-shot retry → bounded drain-loop (handles multiple simultaneously-dead
  connections)
  - is_closed()-only → is_closed() + connection-fatal SqlState detection (handles
  pg_terminate_backend / shutdown / failover, not just socket drops)
  - Same protection extended to SQL Server
  - Both regression tests now genuinely reproduce the failure

  Really appreciate the PR — it was the right call and a clean foundation to build
  on.

PR #184, opened 2026-06-19
test(tags): nested cftransaction blocks must run (savepoint semantics)

bpamiri:
Adds a cross-engine test (red on RustCFML, green on Lucee/ACF/BoxLang) for nested `transaction { ... }` blocks.

## Gap

A `transaction { ... }` block nested inside another `transaction { ... }` block runs to completion on Lucee/ACF/BoxLang — the inner block becomes a **savepoint** on the already-open outer transaction. RustCFML throws at the inner block:

```
cftransaction: nested transactions are not supported
```

## Why it matters

Wheels relies on nested transactions heavily: model `save()`/`create()` and the migrator wrap their work in a transaction, and application code routinely opens its own transaction around several saves — producing a transaction inside a transaction. The Wheels core test suite hits this in **84 specs** (green on Lucee); on RustCFML every model `create()` performed inside an outer application transaction aborts here.

## Repro (verified on v0.228.0)

```cfml
transaction {
    transaction {
        // inner block — savepoint on Lucee/ACF/BoxLang
    }
}
```

- **RustCFML 0.228.0:** throws `cftransaction: nested transactions are not supported`
- **Lucee 5.4 / ACF / BoxLang:** runs to completion (empty blocks need no datasource)

## Test

`tests/tags/test_nested_transaction.cfm` — asserts the inner and outer blocks both run to completion, with a bare single-`transaction{}` block as the in-suite control. Registered in `tests/runner.cfm` beside the other transaction tags tests.

### DIFF
--- a/tests/runner.cfm
+++ b/tests/runner.cfm
@@ -512,6 +512,12 @@ include "harness.cfm";
 <!--- above (which has a body). A bare transaction{} block is the in-suite --->
 <!--- control. --->
 <cf_runtest file="tags/test_transaction_action_statement.cfm">
+<!--- - nested_transaction: a transaction{} block nested inside another --->
+<!--- transaction{} block. Lucee/ACF/BoxLang run the inner as a savepoint and --->
+<!--- complete; RustCFML throws "nested transactions are not supported". Wheels --->
+<!--- model save()/create() inside an outer app transaction hits this (84 specs --->
+<!--- in the core suite). --->
+<cf_runtest file="tags/test_nested_transaction.cfm">
 <!--- - component_declaration_attributes: follow-on to component_soft_keyword. --->
 <!--- Component-header metadata attributes are order-independent and may be --->
 <!--- written quoted or unquoted on Lucee/ACF/BoxLang. Two shapes the Wheels --->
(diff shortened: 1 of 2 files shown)


alexskinner:
Thanks @bpamiri — implemented in **v0.231.0** (commit `9ab3c0f`).

A `transaction { }` block nested inside another no longer throws — it now becomes a **SAVEPOINT** on the already-open outer transaction, matching Lucee/ACF/BoxLang:

- The VM tracks transaction nesting depth plus a per-level savepoint-name stack. The outermost block begins/commits/rolls-back as before; a nested block issues `SAVEPOINT` on start, `RELEASE` on commit, and `ROLLBACK TO` on rollback, leaving the outer transaction open.
- Empty / pending (no-datasource) nested blocks are no-ops, same rationale as a bare `transaction{}` — so your test's inner and outer blocks both run to completion.
- New per-driver savepoint SQL: ANSI `SAVEPOINT`/`RELEASE SAVEPOINT`/`ROLLBACK TO SAVEPOINT` for SQLite/MySQL/PostgreSQL; `SAVE TRANSACTION` + `ROLLBACK TRANSACTION` for SQL Server (release is implicit at commit there).

Your `tests/tags/test_nested_transaction.cfm` passes 3/3 (CLI + serve mode cold & warm). Closing as resolved.

Issue #181, opened 2026-06-19
Lexer: lone unescaped \`#\` in a double-quoted string scans to EOF — misleading "Expected RParen, found Eof" / silent empty output


alexskinner:
### Summary

An **unbalanced (lone) `#` inside a double-quoted cfscript string literal** is mishandled by the lexer. Instead of being reported at the offending line, the `#` is treated as the start of an interpolation and the lexer scans the **rest of the source file** looking for a closing `#`. The result is a misleading parse error pointing at **EOF** (or, in some shapes, silently empty output) — with no indication of where the real problem is.

This is a usability/diagnostics bug: the error message sends you to the wrong line. It bit me while writing a test whose `suiteBegin("… (GitHub #180)")` label contained a literal `#`.

### Minimal repro

```cfc
// hashbug.cfm
<cfscript>
foo("a label with a lone hash (GitHub #180)");
bar(1, 2);
</cfscript>
```

**Actual (rustcfml v0.224.0):**
```
Runtime Error: Parse error in 'hashbug.cfm' [line 5, col 4]: Expected RParen, found Eof
```
Line 5 is EOF; the real issue is the `#` on line 2. The lexer consumed everything after `#` (including `180)");`, the `bar(1,2);` call, and the closing `</cfscript>`) as interpolation content, so the `foo(` call never sees its closing paren.

Escaping the hash (`##180`) parses correctly:
```cfc
foo("a label with an escaped hash (GitHub ##180)");   // parses fine
```

In other shapes (e.g. a top-level `x = "GitHub #180 issue";`) the same root cause surfaces as **silent empty output** (exit 0, nothing printed) rather than an error.

### Expected

Lucee/ACF reject a lone unescaped `#` in a string with a clear compile error **at the offending line/column** (e.g. "invalid construct / unterminated `#` in string"). RustCFML should likewise:
- detect an unterminated `#`-interpolation that runs to EOF, and
- report it at the location of the opening `#`, not at EOF — and never silently swallow it.

### Notes

- Repro confirmed on `main` @ v0.224.0.
- The lexer/string-interpolation scanner needs a bound: if a `#` opens an interpolation and the string literal ends (closing quote) before a matching `#`, that's an error at the opening `#` — it must not run past the string into the rest of the file.

alexskinner:
Fixed in v0.225.0 (commit 4892d1f).

[...]

PR #189, opened 2026-06-19
[codex] Handle same-quote strings inside interpolation

Blute:
## Summary

Fixes a lexer regression where a quoted string literal inside a `#...#` interpolation expression could be mistaken for the end of the outer string when both strings use the same quote character.

This keeps the bounded-interpolation behavior from Issue #181 for genuinely lone hashes, but allows valid CFML such as:

```cfml
"value=#flag ? "YES" : ""#"
"statement=#flag ? "UNIQUE" : ""# INDEX #name#"
'enabled=#flag ? 'YES' : ''#'
```

## Root cause

The interpolation scanner currently treats the outer string's quote character as the end of the string whenever it appears at interpolation paren-depth `<= 0`. That correctly catches a lone hash such as `"GitHub #181"`, but it also rejects valid expression syntax where the interpolation expression itself contains a string literal.

The scanner now recognizes quoted string literals inside interpolation expressions before applying the outer-string boundary check. This prevents nested string content from changing interpolation depth or prematurely ending the outer string.

## Covered scenarios

- double-quoted outer string with double-quoted ternary branches
- empty-string ternary branch
- interpolated string inside a struct literal value
- a following `#name#` interpolation segment after the same-quote expression
- doubled quote escapes inside nested strings
- keyword operators followed by same-quote string literals (`EQ "active"`)
- single-quoted outer string with single-quoted ternary branches
- nested string literals containing escaped `##`

## Validation

- `cargo test -p cfml-compiler`
- `cargo run -- --code 'include "tests/harness.cfm"; include "tests/types/test_string_interpolation_nested_strings.cfm";'`
  - `PASS | Type: string interpolation nested string literals | 7/7 passed`
- `cargo run -- tests/runner.cfm | rg 'SUMMARY|FAIL \\||ERROR'`
  - `SUMMARY: 4271/4271 passed across 529 suites`
- Targeted Lucee 7 check using disposable `lucee/lucee:7.0.3.43-light-nginx` container
  - `SUMMARY: 7/7 passed across 1 suites`

### DIFF
--- a/crates/cfml-compiler/src/lexer.rs
+++ b/crates/cfml-compiler/src/lexer.rs
@@ -304,13 +304,40 @@ impl Lexer {
                     let mut expr_str = String::new();
                     let mut depth = 0;
                     // Scan to the matching (depth-0) '#'. A lone, unbalanced '#'
-                    // must NOT run past the end of the string literal: if we reach
-                    // the closing quote (at depth <= 0, i.e. not inside a nested
-                    // call) the interpolation was never terminated. Report it at
-                    // the opening '#', not at some far-off EOF (GitHub #181).
+                    // must NOT run past the end of the string literal. Quoted
+                    // literals inside the interpolation expression are still part
+                    // of that expression, even when they use the same quote as the
+                    // outer string (for example: "#flag ? "YES" : ""#").
                     while !self.is_at_end() && !(self.current() == '#' && depth == 0) {
-                        if self.current() == '(' { depth += 1; }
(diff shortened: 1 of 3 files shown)


alexskinner:
Landed in v0.230.0 (4dad054). Applied manually rather than via the diff — `main` had since added `[`/`]` depth tracking that the PR's base lacked, so the nested-string consumption was merged on top of that while preserving the existing bracket-subscript behaviour. Your test file and a lexer unit test are included; full suite 4288/4288 + wasm green. Thanks @Blute!

PR #190, opened 2026-06-21
Match Lucee cfcookie Path and SameSite attributes

Blute:
## Summary

- Default omitted `<cfcookie path>` to `Path=/`, matching Lucee browser scope behavior.
- Emit `SameSite=<value>` from the `<cfcookie samesite>` attribute.
- Preserve explicit `path`, existing `Secure`, and existing `HttpOnly` handling.

## Minimal CFML shape

```cfml
<cfcookie name="x" value="1" samesite="Lax">
```

When emitted from a non-root request path, Lucee sends the cookie with `Path=/; SameSite=Lax`. RustCFML previously omitted both attributes, leaving browsers to scope the cookie to the request directory and silently dropping the SameSite policy.

## Root cause

The VM `__cfcookie` handler only emitted `Path` when the `path` attribute was explicitly supplied, and it had no `samesite` branch. The tag parser already passes arbitrary attributes through to `__cfcookie`, so this is fixed at the VM cookie-header emission point.

## Covered scenarios

- Omitted `path` emits `Path=/`.
- Explicit `path` still wins.
- `samesite="Lax"`, `Strict`, and `None` are emitted.
- `samesite="#expression#"` is evaluated before emission.
- Omitted `samesite` does not invent a SameSite attribute.
- Served CFML regression checks real `Set-Cookie` headers and confirms `Secure`/`HttpOnly` still emit.

## Validation

- `cargo test -p cfml-vm --test cfcookie_path_samesite` — 6/6 passed.
- Served RustCFML targeted CFML probe for `tests/tags/test_cfcookie_path_samesite.cfm` — 12/12 passed.
- `cargo run -q -p rustcfml-cli --bin rustcfml -- tests/runner.cfm` — 4308/4308 passed across 539 suites.
- Targeted Lucee validation via `targeted-lucee-test.sh tests/tags/test_cfcookie_path_samesite.cfm` — 12/12 passed.
- `git diff --cached --check` — passed.
- `rustfmt --check --edition 2021 crates/cfml-vm/tests/cfcookie_path_samesite.rs` — passed.

### DIFF
--- a/crates/cfml-vm/src/lib.rs
+++ b/crates/cfml-vm/src/lib.rs
@@ -253,6 +253,20 @@ fn format_cookie_expires(raw: &str) -> String {
     dt.format(FMT).to_string()
 }
 
+fn format_cookie_samesite(raw: &str) -> Option<String> {
+    let trimmed = raw.trim();
+    if trimmed.is_empty() {
+        return None;
+    }
+
+    Some(match trimmed.to_ascii_lowercase().as_str() {
+        "lax" => "Lax".to_string(),
+        "strict" => "Strict".to_string(),
+        "none" => "None".to_string(),
+        _ => trimmed.to_string(),
+    })
+}
+
 /// A cached compiled bytecode program with its source file modification time.
 pub struct CachedProgram {
     pub program: BytecodeProgram,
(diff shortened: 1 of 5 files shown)


alexskinner:
Thanks @Blute! Shipped to `main` as **v0.259.0** (8e61b04).

Reimplemented from your branch (tests-first, then independent fix): `__cfcookie` now defaults an omitted `path` to `Path=/` (site-wide rather than scoped to the request directory) and emits a normalized `SameSite` attribute (Lax/Strict/None case-corrected, other values passed through) when `samesite` is supplied. Explicit `path`, `Secure` and `HttpOnly` handling are unchanged. Your Rust regression suite (6 cases) and CFML serve-mode test (12 assertions) are included; `Set-Cookie` headers verified directly and cross-checked against Lucee.

PR #191, opened 2026-06-22
Preserve cfhttp binary responses

Blute:
## Summary

RustCFML currently decodes `cfhttp` response bodies as text even when `getAsBinary` is requested, and serve mode stringifies binary `cfcontent` payloads instead of writing raw bytes. That breaks CFML proxy patterns where a binary resource is fetched with `cfhttp getAsBinary="yes"` and returned with `cfcontent`.

## Minimal CFML shape

```cfml
<cfhttp url="#url#" method="GET" result="r" getAsBinary="yes">
<cfcontent type="application/octet-stream" variable="#r.fileContent#" reset="true">
```

## Root cause

- The `<cfhttp>` tag parser did not include `getAsBinary` in the generated options struct.
- `fn_cfhttp` always consumed response bodies through `into_string()`, so arbitrary bytes were decoded as text.
- Serve-mode `cfcontent` responses converted `CfmlValue::Binary` through `as_string()`, producing the diagnostic placeholder instead of the original bytes.

## Covered scenarios

- `cfhttp getAsBinary` returns binary `fileContent` for byte payloads.
- Both success and non-2xx `cfhttp` responses preserve binary `fileContent` when requested.
- Served `cfcontent variable=<binary>` responses emit raw bytes.
- Lucee-compatible truthy values for `getAsBinary` include `true`, `yes`, `1`, `binary`, and `always`.

## Validation

- `cargo check -p rustcfml-cli`: passed with the existing `compile_expression_static` dead-code warning in `cfml-codegen`.
- Served RustCFML full runner: `PASS | cfhttp binary response | 4/4 passed`; `SUMMARY: 4387/4387 passed across 540 suites`.
- Targeted Lucee validation: `PASS | cfhttp binary response | 4/4 passed`; `SUMMARY: 4/4 passed across 1 suites`.

No known validation gaps.

### DIFF
--- a/crates/cfml-compiler/src/tag_parser.rs
+++ b/crates/cfml-compiler/src/tag_parser.rs
@@ -888,7 +888,18 @@ fn parse_cf_tag(chars: &[char], start: usize, len: usize, imports: &mut std::col
             // format_attr_value emits literal segments quoted and only evaluates
             // the #...# parts; strip_hashes would collapse the whole value into a
             // bare (mis-parsed) expression like `baseUrlpath`.
-            for key in ["url", "method", "timeout", "charset", "username", "password", "useragent", "proxyserver", "multipart"] {
+            for key in [
+                "url",
+                "method",
+                "timeout",
+                "charset",
+                "username",
+                "password",
+                "useragent",
+                "proxyserver",
+                "multipart",
+                "getasbinary",
+            ] {
                 if let Some(v) = attrs.get(key) {
                     opts.push(format!("{}: {}", key, format_attr_value(v, quoted.contains(key))));
                 }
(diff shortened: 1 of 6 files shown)


alexskinner:
Thanks @Blute! Shipped to `main` as **v0.258.0** (065ccbf).

Reimplemented from your branch (tests-first, then independent fix): `getAsBinary` now flows through the cfhttp tag options; `fn_cfhttp` reads the response body as raw bytes (instead of `into_string`) when binary is requested, on both 2xx and error paths; and serve-mode `build_success_response` emits `CfmlValue::Binary` cfcontent payloads as raw bytes rather than stringifying them into the diagnostic placeholder. Your CFML tests are included; verified in serve mode (exact bytes round-trip) and cross-checked against Lucee.

PR #193, opened 2026-06-22
Preserve cfquery SQL line comments

Blute:
## Summary

`<cfquery>` currently flattens SQL body newlines while lowering tag syntax to `queryExecute(...)`. That changes the semantics of SQL line comments: a PostgreSQL/QoQ `--` comment that should end at the next newline instead consumes the rest of the statement after newline flattening.

## Minimal CFML shape

```cfml
<cfquery name="q" dbtype="query">
    SELECT name
    -- comment
    FROM src
    WHERE id = #targetId#
</cfquery>
```

Lucee preserves the newline after `--`, so the `FROM` and `WHERE` clauses remain part of the SQL statement.

## Root cause

`process_sql_hashes()` trimmed the `<cfquery>` body and replaced every LF with a space before emitting the generated CFScript string. That made the generated SQL effectively one line:

```sql
SELECT name -- comment FROM src WHERE id = ...
```

This PR keeps line breaks intact, normalizing CRLF and bare CR to LF instead of flattening them.

## Covered scenarios

- Static `<cfquery>` SQL with `--` line comments.
- Interpolated `<cfquery>` SQL with `#...#` expressions after a `--` line comment.
- Query-of-Queries coverage keeps the regression self-contained and cross-engine.
- A focused compiler unit pins the tag-lowering output so the generated script keeps the line break.

## Validation

- `cargo test -p cfml-compiler test_cfquery_preserves_sql_line_comment_newlines -- --nocapture`: passed.
- Focused RustCFML CFML test: `PASS | cfquery SQL line comments | 5/5 passed`; `SUMMARY: 5/5 passed across 1 suites`.
- Full RustCFML runner: new test passed; final summary `4380/4380 passed across 544 suites`. The run also emitted an unrelated `ERROR | stdlib/test_cfhttp.cfm | Invalid JSON: EOF while parsing a value at line 1 column 0` while still reporting the all-passed summary.
- Targeted Lucee validation: `PASS | cfquery SQL line comments | 5/5 passed`; `SUMMARY: 5/5 passed across 1 suites`.
- `cargo check -p rustcfml-cli`: passed with the existing `compile_expression_static` dead-code warning in `cfml-codegen`.

## Formatting note

`cargo fmt -p cfml-compiler -- --check` was attempted, but it reports broad pre-existing formatting drift across the package. I did not run formatter to avoid unrelated churn; the changed hunks pass `git diff --check`.

### DIFF
--- a/crates/cfml-compiler/src/tag_parser.rs
+++ b/crates/cfml-compiler/src/tag_parser.rs
@@ -3180,13 +3180,15 @@ fn scan_cfqueryparam_tags(sql_body: &str) -> (String, Vec<CfQueryParam>) {
 /// Converts #var# to string concatenation: `"..." & var & "..."`
 /// Returns a script expression that builds the final SQL string.
 fn process_sql_hashes(sql: &str) -> String {
-    let sql = sql.trim().replace('\n', " ").replace('\r', "");
+    let sql = sql.trim().replace("\r\n", "\n").replace('\r', "\n");
 
     if !sql.contains('#') {
         // No hash expressions — simple string literal. CFML escapes an embedded
         // double-quote by DOUBLING it (""), not with a backslash; SQL bodies use
         // double-quoted identifiers (AS "where"), so a backslash escape would
-        // terminate the string early and fail to parse.
+        // terminate the string early and fail to parse. Preserve line breaks:
+        // SQL `--` comments are line-scoped, and flattening newlines changes
+        // the statement semantics.
         return format!("\"{}\"", sql.replace('"', "\"\""));
     }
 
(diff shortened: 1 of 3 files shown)


alexskinner:
Thanks @Blute! Shipped to `main` as **v0.257.0** (8826533).

Reimplemented from your branch (tests-first, then independent fix): `process_sql_hashes` no longer flattens newlines to spaces — CRLF/CR are normalized to LF and line breaks preserved, so a line-scoped `--` SQL comment ends at the newline instead of swallowing the rest of the query. Your CFML regression test and a compiler unit test are both included. Validated across the full workspace + CLI/serve runner + wasm builds, and cross-checked against Lucee.

PR #196, opened 2026-06-22
[codex] Add regression for custom tag caller scope from CFC method

Blute:
## Summary

This draft PR adds a generic Lucee compatibility regression test for custom tag `caller` scope when a body-mode custom tag is invoked from inside a CFC method.

It intentionally contains only tests and does not propose an implementation change.

## Minimal Shape

```cfml
<cfcomponent>
    <cffunction name="render">
        <cfset payload = { label: "method-scope" } />
        <cfmodule template="customtags/cfc_caller_probe.cfm">BODY</cfmodule>
    </cffunction>
</cfcomponent>
```

The custom tag's end phase reads `caller.payload.label` and `thisTag.generatedContent`.

## Expected vs Actual

On Lucee, the custom tag sees the CFC method-scope variable through `caller` and renders:

```text
caller:method-scope|body:BODY
```

On RustCFML v0.266.0 / current `main`, the body content is delivered correctly, but `caller.payload` is missing:

```text
caller:missing|body:BODY
```

## Covered Scenario

- Body-mode `cfmodule` invoked from a CFC method.
- End-phase custom tag code reading a caller-scope variable.
- Body generated content still delivered through `thisTag.generatedContent`.

## Validation

- `cargo build` passed.
- `./target/debug/rustcfml tests/runner.cfm` fails as expected for the new regression:
  - `Tags: custom tag caller scope from CFC methods | 0/1 passed`
  - expected `caller:method-scope|body:BODY`, got `caller:missing|body:BODY`
  - overall `SUMMARY: 4542/4543 passed across 554 suites`
- `/Users/mat/dev/hello/.agents/skills/rustcfml-upstream/scripts/targeted-lucee-test.sh tests/tags/test_custom_tag_caller_cfc_method.cfm` passed:
  - `SUMMARY: 1/1 passed across 1 suites`

### DIFF
--- a/tests/runner.cfm
+++ b/tests/runner.cfm
@@ -291,6 +291,7 @@ include "harness.cfm";
 <cf_runtest file="tags/test_tags_cfhtmlhead_body.cfm">
 <cf_runtest file="tags/test_tags_cfexit.cfm">
 <cf_runtest file="tags/test_tags_customtag.cfm">
+<cf_runtest file="tags/test_custom_tag_caller_cfc_method.cfm">
 <cf_runtest file="tags/test_custom_tag_attribute_collection.cfm">
 <cf_runtest file="tags/test_tags_customtag_lifecycle.cfm">
 <cf_runtest file="tags/test_tags_buffer_recovery.cfm">
--- a/tests/tags/CustomTagCallerCfcFixture.cfc
+++ b/tests/tags/CustomTagCallerCfcFixture.cfc
@@ -1,0 +1,7 @@
+<cfcomponent output="false">
+    <cffunction name="render" returntype="string" output="false">
+        <cfset payload = { label: "method-scope" } />
+        <cfsavecontent variable="local.out"><cfmodule template="customtags/cfc_caller_probe.cfm">BODY</cfmodule></cfsavecontent>
+        <cfreturn trim(local.out) />
+    </cffunction>
+</cfcomponent>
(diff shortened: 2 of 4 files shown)


noursecode:
help me understand something. Why would you want to support cfmodule or custom tags? Other than perhaps legacy code using it? Seems that calling CFC methods would be cleaner, and more performant. To keep RustCFML relative, consider following the architecture of other popular languages, and toss the use of of tags in logic. If you do, RustCFML might be more inviting to younger devs.

alexskinner:
Fixed in **v0.283.0** (`e44e064`, on `main`). Thanks for the precise regression test, @Blute — it pinpointed this exactly.

**Root cause:** the `__cfcustomtag*` handlers built the custom tag's `caller` scope from the invoking frame's raw locals. When the tag is invoked from inside a CFC method, an unscoped `<cfset payload = ...>` lands in the component's `__variables` sub-scope, but `caller` is supposed to *be* the calling template's *variables* scope — so `caller.payload` was looking one level too high and came back missing.

**Fix:** new `caller_scope_from_locals()` helper surfaces the `__variables` contents as the `caller` scope when present (CFC-method case), falling back to the locals frame at page level (page case unchanged). Writeback diffs against the same view and routes through `scope_aware_store`.

Your test now passes:
```
PASS | Tags: custom tag caller scope from CFC methods | 1/1 passed
SUMMARY: 4641/4641 passed across 574 suites
```
Also green: `cargo test --workspace` (incl. JIT 76/76) and the wasm32 build.

One related edge that's *out of scope* here and remains a known limitation (verified present both before and after this change, so not a regression): writing a **new** key via `caller.newKey = v` from the end phase does not yet propagate back into the CFC method's variables scope. Reading — what this PR reported — is fully fixed. Closing as resolved.

PR #198, opened 2026-06-23
test(core): closure must capture enclosing var-scoped function expressions

bpamiri:
Adds a cross-engine test (red on RustCFML, green on Lucee/ACF/BoxLang) for closure capture of an enclosing `var`-scoped **function expression**.

## Gap

A closure captures the `var`-scoped variables of its enclosing function. On Lucee/ACF/BoxLang that includes function expressions — a helper declared `var fn = function(){...}` can be called by bare name from inside a nested closure. RustCFML captures plain `var` values (string/number/struct) but **not** a `var`-scoped function expression: a bare call from inside the closure throws `Variable fn is undefined` (and even `isClosure(fn)` / a bare read throws).

## Repro (verified on v0.273.0)

```cfml
function buildScope() {
    var strVal = "captured-string";
    var fnVal  = function (x) { return "fn:" & arguments.x; };
    var out = {};
    var inner = () => {
        out.str = strVal;      // OK on RustCFML — value captured
        out.fn  = fnVal("ok"); // RustCFML: "Variable fnVal is undefined"
    };
    inner();
    return out;
}
```

Same nested closure, same enclosing `var` scope:

| Captured var | RustCFML 0.273.0 | Lucee 5.4 / ACF / BoxLang |
|---|---|---|
| `strVal` (string) | `"captured-string"` ✅ | ✅ |
| `fnVal` (function expr) | **`Variable fnVal is undefined`** 🔴 | `"fn:ok"` ✅ |

So plain-value capture already works; only function-expression capture is missing.

## Why it matters

This is the standard test-helper pattern — define a helper at the top of a specs `run()`, call it from inside `it()`/`describe()` closures. The Wheels core suite uses it widely and every such call errors on RustCFML (~21 spec failures across `migrator` and `packages`):

- `vendor/wheels/tests/specs/migrator/MigratorInfoSpec.cfc` — `var insertOrphan = function(version){...}` called from nested `it()` closures (12 failures)
- `vendor/wheels/tests/specs/packages/*` — `var $withEnv`, `var $freshCache` (9 failures)

## Test

`tests/core/test_closure_captures_local_function.cfm` — control assertion (a `var`-scoped string IS captured) passes on RustCFML, isolating function-expression capture specifically. Registered in `tests/runner.cfm` beside the other closure tests.

### DIFF
--- a/tests/core/test_closure_captures_local_function.cfm
+++ b/tests/core/test_closure_captures_local_function.cfm
@@ -1,0 +1,50 @@
+<cfscript>
+suiteBegin("Closure captures enclosing var-scoped function");
+
+// ============================================================
+// Gap surfaced laddering the Wheels framework test suite on RustCFML 0.273.0
+// ============================================================
+// A closure captures the var-scoped variables of its enclosing function. On
+// Lucee/ACF/BoxLang that includes var-scoped FUNCTION EXPRESSIONS: a helper
+// declared `var fn = function(){...}` in the outer function can be called by
+// bare name from inside a nested closure. RustCFML captures plain var VALUES
+// (strings, numbers, structs) correctly, but a var-scoped function expression
+// is NOT visible inside the closure — a bare call throws
+// "Variable 'fn' is undefined", and even isClosure(fn) / a bare read throws.
+//
+// This is the standard test-helper pattern: define a helper at the top of a
+// spec's run() and call it from inside the it()/describe() closures. The Wheels
+// core suite uses it widely (e.g. MigratorInfoSpec `var insertOrphan = ...`
+// called from nested it() closures; packages specs `var $withEnv`, `$freshCache`).
+// On RustCFML every such call errors.
+//
+// catch-body locals don't persist on every engine, so outcomes are recorded in
+// a struct FIELD.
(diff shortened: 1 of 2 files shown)


alexskinner:
Fixed in v0.278.0 (e45a253). Closures now capture an enclosing function's `var`-scoped function expressions, so a helper declared `var fn = function(){...}` is callable by bare name from inside a nested closure.

Root cause was two blanket `Function`-value exclusions in the VM (the `DefineFunction` closure-env seed/sync and the function-entry parent→locals copy) — both there to avoid the `env → Function → captured_scope → env` Arc-cycle leak. The fix captures a var-declared function expression into the closure env with `captured_scope` stripped to `None` (breaks the cycle; the helper still resolves enclosing values via the calling closure's scope), and carries those stripped helpers through at function entry. Named functions and real closures keep `captured_scope: Some(env)` so they're still skipped — no waste, no cycle.

Your test (`tests/core/test_closure_captures_local_function.cfm`) is included in the commit and is green on RustCFML and Lucee 7; the closure-env leak-guard test stays green. Thanks @bpamiri — closing as landed.

Issue #210, opened 2026-06-25
getMetadata().extends omits the superclass functions array — breaks inheritance-chain introspection (inherited lifecycle annotations)


jjannek:
## Summary

`getMetadata( componentInstance ).extends` exists but does **not** contain a `functions` array — the superclass's methods are missing from the inherited metadata. Lucee/ACF include the parent's `functions` (and other members) under `.extends`. This breaks any introspection that walks the inheritance chain via `metadata.extends`.

Engine: **RustCFML v0.298.0**.

## Minimal repro

`Parent.cfc`
```cfml
component {
  function pMethod(){}
}
```

`Child.cfc`
```cfml
component extends="Parent" {
  function cMethod(){}
}
```

`run.cfm`
```cfml
<cfscript>
md = getMetadata( new Child() );
writeOutput( "extends present="           & structKeyExists( md, "extends" ) );
writeOutput( " | extends.functions present=" & ( structKeyExists( md, "extends" ) ? structKeyExists( md.extends, "functions" ) : "n/a" ) );
</cfscript>
```

```
# rustcfml v0.298.0
extends present=true | extends.functions present=false

# Lucee 5
extends present=true | extends.functions present=true   (parent's pMethod listed)
```

## Why it matters

TestBox discovers inherited lifecycle hooks by recursing through `metadata.extends`:

```cfml
// testbox/system/util/Util.cfc :: getAnnotatedMethods()
if ( structKeyExists( arguments.metadata, "functions" ) ) { /* scan own */ }
if ( structKeyExists( arguments.metadata, "extends" ) ) {
    getAnnotatedMethods( arguments.annotation, arguments.metadata.extends );   // recurse into parent
}
```

Because `md.extends.functions` is absent, **annotation-based lifecycle hooks declared in a parent test case never fire** (`@beforeAll`/`@afterAll`/`@beforeEach`/`@aroundEach`). Confirmed via TestBox's own `BDDLifecycleAnnotationsTest` (which extends `ExampleParentTestCase`): the hook counter is `0`/`1` instead of the expected `4`/`8` (Lucee-green). More broadly, any framework that introspects an inheritance chain via `getMetadata().extends.functions` (ColdBox, Preside) is affected.

## Expected

`getMetadata( instance ).extends` includes the parent's `functions` (and the rest of its metadata) recursively, matching Lucee/ACF — so inheritance-chain introspection works.

alexskinner:
[...]

PR #212, opened 2026-06-26
test(oop): getMetadata self-name in ctor + ancestor functions/path on the extends chain

bpamiri:
## getMetadata fidelity: ctor-time self name + ancestor `functions`/`path` on the `extends` chain

Two related `getMetadata` gaps surfaced laddering the Wheels framework test suite on **v0.309.0**. Both are pure-CFML reflection semantics, fail on RustCFML, and pass on Lucee 7.0.4.34.

### (A) `GetMetadata(this).name` is `"Anonymous"` during the pseudo-constructor
Read from inside a component's pseudo-constructor body, `GetMetadata(this).name` returns the literal `"Anonymous"` instead of the class name. It resolves correctly *after* construction — so the engine knows the name, it just isn't populated yet when the constructor runs.

Wheels keys its promote-includes memo on this ctor-time name (`Global.cfc`), so the cache lands under `"Anonymous"` — `promoteIncludedGlobalsMemoSpec` fails with `got [Anonymous]` and an extra spurious cache entry.

### (B) `GetMetadata().extends` ancestor nodes miss `functions` and `path`
The leaf component's metadata has `functions` and `path`; its `.extends` ancestor node does not (Lucee/ACF/BoxLang populate every level of the chain). Wheels' inheritance-chain `config()`-shadow scan (`Controller.cfc::$configOverrideSkipsSuper`) walks the ancestors looking for shadowed functions, finds none, and the dev-mode "called config() without super.config()" warning never fires — `configSuperWarningSpec` fails (4 specs).

### Cross-engine results

|---|---|---|
| `GetMetadata(this).name` in ctor | `Anonymous` ❌ | `GmaCtorFixture` ✅ |
| ancestor `.extends.functions` present | `false` ❌ | `true` ✅ |
| ancestor `.extends.path` present | `false` ❌ | `true` ✅ |

### This PR
Adds `tests/oop/test_getmetadata_fidelity.cfm` (+ fixtures `GmaCtorFixture`, `GmaBaseFixture`, `GmaSubFixture`) and registers it. Validated **red on the v0.309.0 binary** (0/3) and **green on Lucee 7.0.4.34** (3/3).

### DIFF
--- a/tests/oop/GmaBaseFixture.cfc
+++ b/tests/oop/GmaBaseFixture.cfc
@@ -1,0 +1,3 @@
+component {
+	public function baseFn() {}
+}
--- a/tests/oop/GmaCtorFixture.cfc
+++ b/tests/oop/GmaCtorFixture.cfc
@@ -1,0 +1,7 @@
+component {
+	// Read the component's own metadata name DURING the pseudo-constructor.
+	variables.nameAtCtor = getMetadata(this).name;
+	public function getNameAtCtor() {
+		return variables.nameAtCtor;
+	}
+}
--- a/tests/oop/GmaSubFixture.cfc
+++ b/tests/oop/GmaSubFixture.cfc
@@ -1,0 +1,3 @@
+component extends="GmaBaseFixture" {
+	public function subFn() {}
+}
--- a/tests/oop/test_getmetadata_fidelity.cfm
+++ b/tests/oop/test_getmetadata_fidelity.cfm
@@ -1,0 +1,31 @@
+<cfscript>
+suiteBegin("getMetadata fidelity (ctor-time self name + ancestor functions/path)");
+
+// ============================================================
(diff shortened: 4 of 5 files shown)


alexskinner:
Fixed in v0.323.0 (f252e57). Gap (A) — `getMetadata(this).name` returning "Anonymous" during the pseudo-constructor — is fixed: the `this` struct built by the body bakes in the parser placeholder `__name = "Anonymous"` and was only renamed to the real class name AFTER the body returned. The VM now pushes the dotted class name onto a `constructing_component_names` stack before running the body, and getMetadata stamps it onto the in-construction component, so a ctor-time read returns the real name (the post-body rename then no-ops). Gap (B) — ancestor `.extends` nodes carrying `functions[]`/`path` — was already fixed in v0.319.0 (#210). Your test was incorporated as tests/oop/test_getmetadata_fidelity.cfm + fixtures (branch was on a stale main, so only the new files were taken). Thanks @bpamiri!

PR #213, opened 2026-06-26
test(stdlib): serializeJSON must escape control characters per RFC 8259

bpamiri:
## serializeJSON must escape control characters (U+0000–U+001F) per RFC 8259

Surfaced laddering the Wheels framework test suite on **v0.309.0** (`renderingSpec` BEL-byte cases).

RFC 8259 §7 requires control characters U+0000–U+001F to be escaped in JSON strings as `\uXXXX`. RustCFML's `serializeJSON` emits the **raw control byte** instead, so its **own** `deserializeJSON` then rejects the output:

```
Invalid JSON: control character (-) found while parsing a string at line 1 column 10
```

The serializer is self-inconsistent — it produces JSON its own parser refuses, so any serialize→deserialize round-trip of a value containing a control char (e.g. `chr(7)` BEL) fails. `deserializeJSON` is already correct (it accepts the `\uXXXX` escape and rejects raw control bytes per spec); **the fix belongs in the serializer's string escaper.**

### Cross-engine results

```cfm
jsonStr = serializeJSON({msg = "a" & chr(7) & "b"});
deserializeJSON(jsonStr).msg;   // round-trip
```

|---|---|---|
| serialize emits | raw `0x07` byte ❌ | `` escape ✅ |
| round-trip | throws (own parser rejects) ❌ | `len = 3` ✅ |

### This PR
Adds `tests/stdlib/test_serializejson_control_chars.cfm` and registers it. Validated **red on the v0.309.0 binary** (0/2) and **green on Lucee 7.0.4.34** (2/2).


alexskinner:
Fixed in v0.324.0 (26cb260). serializeJSON emitted raw control bytes (U+0000–U+001F) in string values, so its own deserializeJSON (and any RFC 8259 parser) rejected the output. Added a `json_escape_str` helper using the standard short escapes for `\" \\ \b \f \n \r \t` and `\u00XX` for every other control char; routed string values, struct keys, query column names, and the NativeObject marker through it (this also fixes backslashes in keys/column names that the old `"`-only replace missed). Round-trips correctly now. Your test was incorporated as tests/stdlib/test_serializejson_control_chars.cfm. Thanks @bpamiri!

PR #214, opened 2026-06-26
test(qoq): queryExecute maxrows option + list:true IN-clause expansion

bpamiri:
## queryExecute `maxrows` option + `list:true` IN-clause expansion

Two `queryExecute` query-binding gaps surfaced laddering the Wheels framework test suite on **v0.309.0** (`DatabaseAdapterSpec` cleanup). Both reproduce in **pure Query-of-Queries** (`dbtype="query"`), so they are engine-wide query-binding gaps — not database-specific.

> Note: the original hypothesis ("DELETE … LIMIT / `SQLITE_ENABLE_UPDATE_DELETE_LIMIT`") was a red herring. `DatabaseAdapter.cfc::cleanup()` does not emit `DELETE … LIMIT`; it bounds a `SELECT` with the `maxrows` option, then `DELETE`s `WHERE id IN (:ids)` using a `list=true` param. Those two primitives are the actual gaps.

### (A) the `maxrows` option is ignored
`queryExecute(sql, {}, {dbtype="query", maxrows=2})` returns all rows instead of capping at 2.

### (B) a `list=true` param does not expand into an IN() value list
`queryExecute("… WHERE id IN (:ids)", {ids={value="1,2,3", list=true}}, …)` matches 0 rows instead of expanding to `IN (1,2,3)`.

### Cross-engine results

|---|---|---|
| `maxrows=2` → recordCount | `5` ❌ | `2` ✅ |
| `list=true` IN() → recordCount | `0` ❌ | `3` ✅ |

### This PR
Adds `tests/qoq/test_queryexecute_maxrows_list.cfm` and registers it. Validated **red on the v0.309.0 binary** (0/2) and **green on Lucee 7.0.4.34** (2/2).

### DIFF
--- a/tests/qoq/test_queryexecute_maxrows_list.cfm
+++ b/tests/qoq/test_queryexecute_maxrows_list.cfm
@@ -1,0 +1,30 @@
+<cfscript>
+suiteBegin("queryExecute maxrows option + list:true IN-clause expansion");
+
+// ============================================================
+// Gaps surfaced laddering the Wheels framework test suite on RustCFML 0.309.0.
+// ============================================================
+// Both reproduce in pure Query-of-Queries (dbtype="query"), so they are
+// engine-wide query-binding gaps, NOT database-specific. Surfaced by
+// DatabaseAdapterSpec's cleanup() (a SELECT bounded by the `maxrows` option,
+// then a DELETE ... WHERE id IN (:ids) using a list=true param). The cluster's
+// "DELETE ... LIMIT" hypothesis was a red herring — these two query primitives
+// are the actual gaps.
+//   (A) the `maxrows` queryExecute option must cap the result set.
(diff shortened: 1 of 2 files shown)


alexskinner:
Fixed in v0.325.0 (426cfe3). Both QoQ (dbtype="query") query-binding gaps are addressed: (A) the `maxrows` queryExecute option is now extracted and caps the result query's rows before any returntype conversion (negative = no cap, the Lucee sentinel); (B) a `list=true` param is split on its delimiter (default ",", honours separator/delimiter) into an array, and the QoQ IN evaluator flattens an array-valued list element so `IN (:ids)` matches any member. NOT IN, numeric coercion, and single-element lists all verified. Your test was incorporated as tests/qoq/test_queryexecute_maxrows_list.cfm. Thanks @bpamiri!

PR #293, opened 2026-08-01
RED test: PostgreSQL queryExecute failures must be catchable as type=database

Blute:
## What

Adds a database test proving a failed `queryExecute` raises a **database-typed** exception, so `catch (database e)` / `<cfcatch type="database">` matches (Lucee/ACF behavior). SQLite leg runs always as the control; the PostgreSQL leg is live-gated on `RUSTCFML_TEST_PG_DS` (same convention as `test_db_null_column_empty_string.cfm`) and skips green when unset.

## Why

The SQLite and MySQL adapters already raise `CfmlError::database` (~30 call sites, e.g. builtins.rs:9139, 9398, 9514). The PostgreSQL adapter is the odd one out: every query failure funnels through `PgRunError::from_postgres` (crates/cfml-stdlib/src/builtins.rs:11607), which wraps it as `CfmlError::runtime` — so a database-typed catch never matches and the error falls through to `catch (any)` with `cfcatch.type = Runtime`. The two param-rewrite errors in pg_sql.rs:88/103 share the blind spot. The VM machinery is already in place: `catch_type_matches` (cfml-vm/src/lib.rs:4109) and the `"database"` mapping used by cfthrow (lib.rs:4612).

Real-world repro: moopa's first-run fallbacks wrap "table may not exist yet" probes in `<cfcatch type="database">`. On Lucee this boots an empty database cleanly; on RustCFML + Postgres (Neon) the catch never matches, so first boot 500s intermittently — we had to widen the catches to `type="any"` as a workaround (moopa `route_registry_store.cfc`, `lib/core.cfc`).

Stretch goal while you're in there: populating `cfcatch.sqlstate` from `e.code()` would complete Lucee parity for database catches.

## Validation

RustCFML **v0.541.0** stock release binary, live Postgres 16 (`RUSTCFML_TEST_PG_DS` set):

```text
FAIL | queryExecute failures are catchable as type=database | 1/3 passed (2 failed)
FAIL: PostgreSQL: missing-table error lands in catch (database e) | expected: [database] | got: [any (cfcatch.type=Runtime)]
FAIL: PostgreSQL: cfcatch.type is database | expected: [database] | got: [runtime]
```

Without `RUSTCFML_TEST_PG_DS` (skip path):

```text
PASS | queryExecute failures are catchable as type=database | 2/2 passed
```

The SQLite control passes today, pinning the existing correct behavior. Test-only PR so the Lucee-compatible exception typing is pinned before implementation.

### DIFF
--- a/tests/database/test_query_error_catch_type_database.cfm
+++ b/tests/database/test_query_error_catch_type_database.cfm
@@ -1,0 +1,64 @@
+<cfscript>
+// A queryExecute failure must raise a DATABASE-typed exception so
+// `catch (database e)` / `<cfcatch type="database">` matches — Lucee/ACF parity.
+//
+// The SQLite and MySQL adapters already raise CfmlError::database, but the
+// PostgreSQL adapter funnels every query failure through
+// PgRunError::from_postgres (crates/cfml-stdlib/src/builtins.rs), which wraps
+// them as CfmlError::runtime — so a database-typed catch never matches and the
+// error falls through to `catch (any)`. (pg_sql.rs's two param-rewrite errors
+// have the same blind spot.)
+//
+// Repro class: framework first-run fallbacks wrap "table may not exist yet"
+// probes in <cfcatch type="database">. On RustCFML + Postgres the catch never
+// matches, so booting against an empty database throws instead of falling back
+// (moopa route_registry_store.cfc / core.cfc had to widen to type="any").
+//
+// The PostgreSQL leg is live-gated (same convention as
+// test_db_null_column_empty_string.cfm) and skips when the env var is unset:
+//   docker run -d -p 55432:5432 -e POSTGRES_PASSWORD=test -e POSTGRES_DB=testdb postgres:16
+//   RUSTCFML_TEST_PG_DS=postgres://postgres:test@127.0.0.1:55432/testdb \
+//     cargo run --features all-databases -- tests/runner.cfm
+
+suiteBegin("queryExecute failures are catchable as type=database");
+
+function envDs(required string varName) {
+	try { return getEnvironmentVariable(arguments.varName, ""); }
+	catch (any e) { return ""; }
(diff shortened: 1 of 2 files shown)


alexskinner:
Fixed and shipped in v0.542.0 (aa8dbe9) — your test is in the tree, thanks Matthew.

Your diagnosis was exactly right. `PgRunError::from_postgres` (builtins.rs:11610) and both positional-parameter errors in pg_sql.rs now raise `CfmlError::database`, so `catch (database e)` / `<cfcatch type="database">` matches on PostgreSQL the way it already did on SQLite/MySQL/MSSQL. Verified RED (1/3) then GREEN against live PostgreSQL 16 with `RUSTCFML_TEST_PG_DS` set.

Auditing the rest of the DB layer while in there turned up a second, wider gap your PR didn't cover: `cftransaction` control failures — pool checkout, BEGIN/COMMIT/ROLLBACK, savepoints — were runtime-typed on **every** engine, not just PostgreSQL (SQLite/MySQL inline, MSSQL via `mssql_txn_control`). All 17 sites are now database-typed. There is no `CfmlError::runtime` left anywhere in the database layer. Your test file gained two cftransaction cases to pin that; it now runs 5/5 with a live PG and 3/3 on the skip path.

MSSQL's query path was already correct, so it needed no change there.

On the stretch goal — `cfcatch.sqlstate` from `e.code()` isn't a one-liner: `CfmlError` is `{message, error_type, stack_trace}` with no structured extras, so it needs a new field threaded through the VM's exception-struct builder plus per-engine plumbing. Worth doing as its own change; happy to take a PR or pick it up.

You should be able to drop the `type="any"` widening in moopa's route_registry_store.cfc / lib/core.cfc.

alexskinner:
Tracked the sqlstate stretch goal as #295, credited to you.

PR #294, opened 2026-08-01
RED test: cfquery tag must forward columnkey and maxrows like queryExecute

Blute:
## What

Adds a database test proving the `<cfquery>` **tag** honours `returntype="struct" columnkey="..."` (row-keyed struct, Lucee behavior) and `maxrows` the same way `queryExecute()` already does. Runs entirely on the in-memory SQLite datasource — no live DB or env gating needed. The three `queryExecute` assertions pass today and pin the working path; the three tag assertions fail.

## Why

Both options are implemented and work through `queryExecute` (columnkey since v0.13.0 / 64be57ee, including the `keyColumn` alias from 72fb6c38; maxrows capping in builtins.rs `apply_query_maxrows`). But the tag lowering forwards only a whitelist of attributes into the options struct — `tag_parser.rs` "cfquery" arm: `datasource / name / result / returntype / dbtype / attributeCollection` — so `columnkey` and `maxrows` are silently dropped. With `returntype="struct"` and no columnkey, the builtin falls into its single-row-map branch and returns a struct keyed by **column names** instead of row key values.

Bisected against release binaries **v0.125.0, v0.126.0, v0.531.0 and v0.541.0** — identical on all four, so this isn't a regression; the tag path has never forwarded these attributes. (64be57ee's end-to-end verification presumably exercised the builtin path.)

Real-world repro: moopa's hub schema-sync reads existing index/FK metadata with `<cfquery returntype="struct" columnkey="name">` (`schemaSync.cfc` `getTableIndexes` / `getTableForeignKeys` / `getTablePrimaryKey`). On RustCFML the "existing objects" struct comes back keyed by the query's column names, so against a perfectly healthy schema the diff page proposes `ALTER TABLE ... DROP CONSTRAINT name / table_name / foreign_key_table / onupdate / ondelete` on every table, `DROP INDEX name / fields / unique / type / indexdef / where`, and re-`CREATE` of every index and FK that already exists.

## Validation

RustCFML **v0.541.0** stock release binary:

```text
FAIL | cfquery tag forwards columnkey and maxrows like queryExecute | 3/6 passed (3 failed)
FAIL: cfquery tag: struct is keyed by the columnkey column's row values | expected: [alpha,beta] | got: [name,v]
FAIL: cfquery tag: row key resolves to its row struct | expected truthy | got: [false]
FAIL: cfquery tag: maxrows caps the resultset | expected: [1] | got: [2]
```

Test-only PR so tag/queryExecute parity is pinned before implementation.

### DIFF
--- a/tests/database/test_cfquery_tag_forwards_columnkey_maxrows.cfm
+++ b/tests/database/test_cfquery_tag_forwards_columnkey_maxrows.cfm
@@ -1,0 +1,53 @@
+<cfscript>
+// The <cfquery> TAG must behave identically to queryExecute() for the same
+// attributes. queryExecute honours returntype="struct" + columnKey (row-keyed
+// struct, Lucee behavior — supported since v0.13.0) and maxrows — but the tag
+// lowering forwards only a whitelist of attributes into the options struct
+// (tag_parser.rs "cfquery" arm: datasource/name/result/returntype/dbtype/
+// attributeCollection), so columnkey and maxrows are silently dropped.
+//
+// With columnkey dropped, returntype="struct" falls into the "single-row map"
+// branch and returns a struct keyed by COLUMN NAMES instead of row key values.
+//
+// Real-world repro: moopa's hub schema-sync reads index/FK metadata with
+// <cfquery returntype="struct" columnkey="name"> (schemaSync.cfc
+// getTableIndexes/getTableForeignKeys). On RustCFML the "existing indexes"
+// struct comes back keyed by the query's column names, so the schema diff
+// offers to DROP CONSTRAINT name/table_name/onupdate/... on every table and
+// re-CREATE every index and FK that already exists.
+//
+// Verified identical on v0.125.0, v0.126.0, v0.531.0 and v0.541.0 release
+// binaries — the tag path has never forwarded these attributes.
+
+suiteBegin("cfquery tag forwards columnkey and maxrows like queryExecute");
+
+memDs = { class: "org.sqlite.JDBC", connectionString: "jdbc:sqlite::memory:" };
+twoRowSql = "SELECT 'alpha' AS name, '1' AS v UNION ALL SELECT 'beta', '2'";
+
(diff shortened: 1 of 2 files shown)


alexskinner:
Fixed and shipped in v0.543.0 (de89409) — your test is in the tree, thanks Matthew.

Your diagnosis was spot on, right down to the line. The `<cfquery>` lowering in `tag_parser.rs` built its options struct from a hard-coded whitelist of six attributes, so `columnkey`, `keycolumn`, `maxrows` and `timeout` were dropped on the tag path while working fine through `queryExecute()`. With `columnkey` gone, `returntype="struct"` fell into the single-row-map branch and came back keyed by column names — exactly the schema-sync "drop every index and FK on a healthy schema" behaviour you hit.

The fix forwards **all** attributes rather than extending the whitelist, canonicalising the spellings the VM intercept and the builtin look up, and emitting in source order (`attr_order`) so codegen stays deterministic — `attrs` is a `HashMap`. Options the engine does not recognise are ignored downstream, as in Lucee.

Worth noting what made this a whitelist bug rather than a missing feature: the script-tag path in `parser.rs` (`query name="..." { ... }`) already forwarded every attribute. So the two `cfquery` spellings disagreed with each other, which is why your bisect found it identical all the way back to v0.125.0 — the tag path was never right, and the builtin-path verification behind 64be57ee could not have caught it.

I added two cases to your test: the `keyColumn` alias with a dynamic `#expr#` value, and a body containing `<cfif>`. That second one matters — control flow in the body switches the lowering to the runtime savecontent branch, which shares the same options struct, so it needed pinning too. RED 3/6 on the v0.542.0 stock binary, GREEN 8/8 with the fix.

One caveat now that the floodgate is open: `cachedwithin` reaches the options struct but the engine has no query cache, so it silently no-ops rather than erroring. Not a regression — it was inert before too, just discarded earlier in the pipeline — but do not read it as working.

You should be able to drop the workaround in moopa's schemaSync.cfc.

PR #296, opened 2026-08-01
RED test: cfloop list with both item and index must populate both (Lucee parity)

Blute:
## What

Adds a core test pinning Lucee's `<cfloop list>` semantics when **both** `item` and `index` are present: `item` = the list element, `index` = the 1-based position. The two single-attribute forms (item-only → element; legacy index-only → element) already match Lucee and are pinned as controls.

## Why

RustCFML currently treats `item` + `index` like the legacy index-only form — the element lands in `index` and `item` is never set, so any read of the item variable throws `Variable '...' is undefined`.

Real-world repro: moopa's hub schema-sync FK comparison —

```cfml
<cfloop list="#structKeyList(code_fk)#" item="code_fk_param" index="i">
    <cfif code_fk[code_fk_param] NEQ (db_foreign_keys[code_fk_name][code_fk_param]?:'')>
```

(`schemaSync.cfc` `compareDatabaseSchema`). The sysadmin schema page 500s with `Variable 'code_fk_param' is undefined` as soon as an existing foreign key is compared. Fun detail: this line was unreachable until **v0.543.0** — your GH #294 fix made the FK metadata struct readable, which let this loop execute on RustCFML for the first time. One layer down the onion.

## Validation

RustCFML **v0.543.0** stock release binary:

```text
FAIL | cfloop list with both item and index | 2/4 passed (2 failed)
FAIL: item+index: item receives each element | expected: [a,b,c] | got: [UNDEF,UNDEF,UNDEF]
FAIL: item+index: index receives the 1-based position | expected: [1,2,3] | got: [a,b,c]
```

Lucee **6.2** direct check (lucee/lucee:6.2 Docker, same probe): `item-only=a,b,c`, `index-only=a,b,c`, both: `item=a,b,c` `index=1,2,3`.

Test-only PR so the Lucee-compatible loop-variable population is pinned before implementation.

### DIFF
--- a/tests/core/test_cfloop_list_item_with_index.cfm
+++ b/tests/core/test_cfloop_list_item_with_index.cfm
@@ -1,0 +1,47 @@
+<!--- <cfloop list> with BOTH item and index must populate both, Lucee-style:
+      item = the list element, index = the 1-based position. RustCFML currently
+      treats the combination like the legacy index-only form — the element
+      lands in `index` and `item` is never set, so any read of the item
+      variable throws "Variable '...' is undefined".
+
+      The two single-attribute forms already match Lucee (item-only → element;
+      legacy index-only → element) and are pinned here as controls.
+
+      Real-world repro: moopa's hub schema-sync FK comparison —
+        <cfloop list="#structKeyList(code_fk)#" item="code_fk_param" index="i">
+      (schemaSync.cfc compareDatabaseSchema). On RustCFML the sysadmin schema
+      page 500s with "Variable 'code_fk_param' is undefined" as soon as an
+      existing foreign key is compared. (Unreachable before v0.543.0 — the
+      cfquery tag columnkey fix (GH #294) made the FK metadata readable, which
+      let this loop execute for the first time.)
+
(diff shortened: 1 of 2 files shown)


alexskinner:
Fixed and shipped in v0.544.0 (5ad0f0c) — your test landed verbatim alongside the fix, thanks.

Root cause was the same shape as #294: two lowerings for the same tag that had drifted. Both treated `item` as a plain alias for `index` on list loops, so the element went into `index` and the item variable was never bound:

- `crates/cfml-compiler/src/tag_parser.rs` — the `<cfloop>` tag preprocessor
- `crates/cfml-compiler/src/parser.rs` — the script `loop …` / `cfloop( … ){}` form

Both now mirror the existing `array=` item+index lowering: hoist `listToArray()` into a temp, count a 1-based position, bind `index` = position and `item` = element. Single-attribute forms (either spelling names the element) are unchanged.

Verified byte-identical against Lucee 7.0.4 across tag form, script paren form, bare `loop` statement form, `delimiters=`, nested loops, and `break`/`continue`. Your test file passes on both engines (4/4). Closing since the commit carries the test.

PR #309, opened 2026-08-02
Debug footer: never render process environment variables

Blute:
## Problem

The debug footer's **Environment variables** section dumps the engine's entire process environment. Deployment env vars routinely carry credentials — datasource passwords, API keys, signing secrets — and the footer's only gate can be the `debugging.urlTrigger` URL key. Anyone holding (or guessing) that key gets the raw database password, even though the CFConfig section already credential-redacts its own rows (e.g. `datasources.*.password` renders as `••••••`). The env dump bypassed that redaction entirely.

## Change

- Remove the Environment variables section from the footer outright. CFConfig (already redacted) and the CLI **Runtime flags** section are kept.
- `render_env_and_flags` → `render_runtime_flags`; comments document why env vars must never render.
- Tests updated: the canonical-order test and section test now assert the env section is **absent**.

`cargo test -p cfml-vm --lib`: 147 passed.

### DIFF
--- a/crates/cfml-vm/src/debug_footer.rs
+++ b/crates/cfml-vm/src/debug_footer.rs
@@ -653,11 +653,13 @@ fn render_html(
     }
 
     // Scopes, in the canonical order set by `gather_debug_scopes`: url, form,
-    // cgi. The deploy-level blocks (cfconfig overrides, then the engine
-    // environment: process env vars + CLI flags) render directly under the cgi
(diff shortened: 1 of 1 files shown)


Blute:
Closing — keeping the env dump as an operator feature; the right control is key hygiene on the deployment side (strong or empty urlTrigger value, production mode refusing presence-only triggers).

PR #311, opened 2026-08-03
Fixes #310

### LINKED ISSUE #310: <cfexit method="loop"> is silently inert: custom tag bodies never re-execute (listings render one row)
## Summary

`<cfexit method="loop">` is accepted but silently does nothing: the custom tag's
body is never re-executed, so the tag runs its body exactly once regardless of
how many times it asks to loop. There is no error and no warning — the page just
renders one iteration's worth of content with HTTP 200.

This is the documented mechanism a custom tag uses to iterate its own body
(cfdocs: *"loop — used only in a custom tag body; re-executes the body of the
custom tag"*), and it is how tag-based CFML frameworks implement row/record
iteration tags. Any such tag currently renders **only its first row**.

Behaviour verified live on a v0.538.0 build; reference Lucee 7.0.4.34 (same on
5.3.10.120). 

### 1. Minimal repro

`_rustpoc_rtags/looptag.cfm` — a tag that asks for three iterations:

```cfml
<cfparam name="request.rloopCount" default="0">
<cfif thisTag.executionMode eq "end">
    <cfset request.rloopCount = request.rloopCount + 1>
    <cfif request.rloopCount lt 3>
        <cfexit method="loop">
    </cfif>
</cfif>
```

page:

```cfml
<cfimport taglib="/_rustpoc_rtags" prefix="r">
<cfset request.rloopCount = 0>
<r:looptag><cfoutput>(BODY)</cfoutput></r:looptag>
<cfoutput>[count]#request.rloopCount#</cfoutput>
```

|           | Lucee 7.0.4            | RustCFML v0.538.0 |
|-----------|------------------------|-------------------|
| output    | `(BODY) (BODY) (BODY)` | `(BODY)`          |
| `[count]` | `3`                    | `1`               |

The body is re-executed once per loop request on Lucee, and each iteration's
output accumulates into the page. On RustCFML the end phase runs again but the
body does not, so only the first iteration's output is ever emitted.

### 2. Source note

The plumbing is largely in place: `last_exit_method` already carries the method
string and its doc comment lists `"loop"` alongside `"exittag"` /
`"exittemplate"` (`crates/cfml-vm/src/lib.rs:1823`), and `__cfcustomtag_start` /
`__cfcustomtag_end` already implement the `exittag` path via `CustomTagState`.
But `"loop"` is never consumed anywhere in `cfml-vm` or `cfml-codegen` —
grepping `main` for it in those crates returns only that one doc comment.

Note that the obvious source-level shortcut is not safe: the tag preprocessor
expands a custom-tag pair into plain script
(`tag_parser.rs`: `__cfcustomtag_start(...); <body> __cfcustomtag_end();`), so
one could wrap the body in a synthetic `while (true) { … }`, but that would
capture a `<cfbreak>` written in the body and intended for the *caller's* own
`<cfloop>` — a behaviour change. Keeping the body as inline bytecode and having
the execution loop rewind to it preserves break/continue binding exactly.

A PR implementing it that way follows.

Yucheng-Yan:
Fixes #310

## What

`<cfexit method="loop">` was accepted but silently inert: it parsed, ran, and
landed in `last_exit_method`, which nothing ever consumed (before this change:
`exittag` 5 consumption sites, `exittemplate` 1, **`loop` 0**). Any custom tag
that iterates its own body therefore rendered only its first row.

The execution loop now records the tag body's instruction index right after
`__cfcustomtag_start` returns, and rewinds to it when the end phase requests
another iteration.

### Why rewind instead of a synthetic loop

The tag preprocessor already expands a pair into flat script
(`__cfcustomtag_start(...); <body> __cfcustomtag_end();`), so the tempting fix is
to emit `while (true) { <body> if (!__cfcustomtag_end()) break; }`. That would
**rebind a `<cfbreak>` written in the body and intended for the caller's own
`<cfloop>`** to the synthetic loop — a silent behaviour change. Keeping the body
as inline bytecode avoids it, and needs **no codegen and no preprocessor change**.

`break`/`continue` behaviour in tag bodies is unchanged by this patch. 

Stack invariant worth flagging for review: an expression statement compiles to
`… Call; Pop`, and `ip` is advanced before the op match, so at the record site
`ip` is that trailing `Pop` and `ip + 1` is the body. The code asserts the `Pop`
is really there (bounds-checked) and **fails fast** if it isn't, after unwinding
the tag state and restoring the outer output buffer — a broken VM/codegen
invariant should not degrade into silently wrong output. On the rewind path the
end call's result is popped explicitly, since we jump over the `Pop` that would
otherwise have consumed it.

## Two pre-existing divergences fixed along the way

Both were required for iterations to work at all, and each ships with its own
assertion:

1. **`thisTag` was rebuilt wholesale at the end phase**, discarding members the
   tag set itself. Lucee keeps them — probe: `state=S0` on Lucee vs **`GONE`** on
   `main`. It is now refreshed **in place** (mutating the existing struct, not
   replacing it), so members survive *and* an alias the tag kept
   (`variables.saved = thisTag`) still observes the current phase rather than a
   stale snapshot. Only `executionMode` / `hasEndTag` / `generatedContent` change.
2. **`last_exit_method` was taken only in start phases**, so a `cfexit` fired in
   an *end* phase stayed set until some unrelated later tag consumed it — and if
   it was `exittag`, that tag's body and end phase were silently skipped. It is
   now taken exactly once per phase boundary with an exhaustive match
   (`loop` / `exittag` / `exittemplate` | none / other). An unrecognised method
   keeps `main`'s current behaviour; it is only consumed so it cannot leak.

## Semantics — measured on Lucee 7.0.4.34, not assumed

| Behaviour | Lucee |
|---|---|
| start phase of a looped tag | runs **once** — loop re-runs body + end phase only |
| body | re-executed per iteration |
| `thisTag` members | persist across iterations (`S0 → S1 → S2`) |
| `thisTag.generatedContent` | per-iteration, never accumulated |
| `attributes`, `executionMode` | stable (`end` every round) |
| page output per iteration | generatedContent, then the end phase's own output |
| loop requested in the **start** phase | throws — this patch raises Lucee's message |
| loop from a **self-closing** tag | allowed; re-runs the (empty) body + end phase |

The self-closing case is a separate engine path (no start/end pair, no rewind)
and is handled in its own arm, with its own tests.

## Tests

Extends the existing `tests/tags/test_tags_cfexit.cfm` suite (already registered
in `runner.cfm`, so no runner change): **3 → 22 assertions**, plus 7 tag
templates in `tests/tags/`.

## Verification

- `Tags: cfexit` **22/22 on RustCFML and 22/22 on Lucee 7.0.4.34** — including
  the page-output assertion, so the expected ordering is Lucee's real behaviour.
- Full CFML suite: **7090/7091 CLI**, **7122/7123 served — dev *and*
  `--production`, cold and warm**. The single failure is a pre-existing
  datasource assertion (`cftransaction: an unopenable datasource is
  database-typed`) that fails identically on an unpatched build of this base.
- `cargo test -p cfml-vm --lib` 147/147, **JIT suite `jit_numeric` 76/76**,
  `wasm32-unknown-unknown` build of `cfml-worker` + `rustcfml-wasm` OK.

### DIFF
--- a/crates/cfml-vm/src/lib.rs
+++ b/crates/cfml-vm/src/lib.rs
@@ -1828,8 +1828,15 @@ pub struct CfmlVirtualMachine {
     pending_exit: Option<String>,
     /// The `cfexit` method carried out of the most recently unwound template
     /// frame. Consumed by the custom-tag handlers to decide exittag (skip body +
-    /// end) vs exittemplate (run body + end normally).
+    /// end) vs exittemplate (run body + end normally) vs loop (re-run the body).
+    /// Taken with a single `.take()` per phase boundary: a method fired in an end
+    /// phase must not leak into an unrelated later tag.
     last_exit_method: Option<String>,
+    /// Set by `__cfcustomtag_start`: the execution loop (which owns `ip`) records
+    /// the body's instruction index into the tag state just pushed.
+    pending_tag_body_start: bool,
+    /// Set by `__cfcustomtag_end` on `cfexit method="loop"`: rewind target.
+    pending_tag_loop: Option<usize>,
     /// Buffered `<cfhtmlhead>` / `<cfhtmlbody>` content, injected into the
     /// response `<head>` / `<body>` (or appended) at output finalization.
     html_head_buffer: String,
@@ -2331,6 +2338,10 @@ struct CustomTagState {
     /// end phase must be skipped, but partial start-phase caller writes still
     /// propagate.
     exit_tag: bool,
+    /// Instruction index of the tag BODY in the *calling* frame — the rewind
+    /// target for `cfexit method="loop"`. The body stays inline bytecode (a jump
+    /// re-runs it) so `cfbreak`/`cfcontinue` in a body keep binding as before.
+    body_start_ip: Option<usize>,
 }
 
 impl CfmlVirtualMachine {
@@ -2424,6 +2435,8 @@ impl CfmlVirtualMachine {
             custom_tag_stack: Vec::new(),
             pending_exit: None,
             last_exit_method: None,
+            pending_tag_body_start: false,
+            pending_tag_loop: None,
             html_head_buffer: String::new(),
             html_body_buffer: String::new(),
             app_function_table: Vec::new(),
@@ -4741,6 +4754,9 @@ impl CfmlVirtualMachine {
         if self.custom_tag_stack.len() > handler.custom_tag_depth {
             self.custom_tag_stack.truncate(handler.custom_tag_depth);
         }
+        // A throw in a tag body must not leave a rewind target for an abandoned tag.
+        self.pending_tag_loop = None;
+        self.pending_tag_body_start = false;
     }
 
     /// Lucee/ACF parity for the `arguments` scope: reading a declared-but-unpassed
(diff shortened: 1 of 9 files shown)


alexskinner:
generally I would prefer for core engine changes with this blast radius to just receive the issue << the pull request is useful if its just to provide the test which illustrates the issue but the actual fix isn't because usually the core has changed a fair amount so its unlikely to be mergeable

alexskinner:
Landed in **v0.563.0** (49e0868), built on this PR's analysis rather than merged directly — so credit where it's due, and an explanation of the differences.

**What this PR got right, and what it saved:**

- The core mechanism. Keeping the body as inline bytecode and rewinding to it is the only approach that works, and the reasoning is correct: `break`/`continue` are resolved at codegen time against the compiler's `loop_stack` (`compiler.rs:93`, `:1738`), so lowering the body into a synthetic `while (true) { … }` really would rebind a `<cfbreak>` written in the body and intended for the caller's own `<cfloop>`. That was worth spelling out and it held up under checking.
- Both incidental divergences you flagged are real. I re-probed them independently against Lucee 7.0.4.34 rather than take them on trust, and got exactly what you described:

  ```
                                        Lucee   RustCFML (before)
  thisTag.mine survives start -> end :  S0      GONE
  alias variables.saved.executionMode:  end     start
  ```

  Both are fixed, in place rather than by rebuild, as you had it.

**What changed in the shipped version:**

1. **The rewind target is resolved at codegen, not inferred at runtime.** This PR records the body index as `ip + 1`, reasoning that an expression statement compiles to `… Call; Pop`. That holds today, and the bounds-check-and-fail-fast guard is a sound way to defend it. But `compile_statement` unconditionally prepends a `LineInfo` to every statement and several peepholes reshape that arm, so the assumption is a layout contract living in the VM dispatch loop where nothing would flag its violation. The shipped version keeps a lexical stack of open pairs in the compiler, records `instructions.len()` after the start call's `Pop`, and emits `TagLoopBack(body_start)` in place of the end call's trailing `Pop`. Same mechanism, target self-describing in the bytecode, no runtime assertion needed.

2. **`<cfexit method="loop">` outside an end phase now throws.** Lucee raises `invalid context for the tag exit, method loop can only be used in the end tag of a custom tag` from `Exit.doEndTag`. Silently ignoring it would have left the same class of bug this PR fixes — an accepted directive that does nothing.

3. **Non-local exit from a tag body is unwound.** This turned out to be a live bug independent of `method="loop"`: a `<cfbreak>` in a tag body skips `__cfcustomtag_end`, so the buffer pushed by the start tag is never popped and the rest of the page writes into it and is silently discarded — truncated output, HTTP 200, no error. Since the whole point of `method="loop"` is bodies that iterate, it made sense to fix in the same change. Codegen emits an `AbandonTagPairs` before a `break`/`continue` that exits open pairs; frame exit covers `return`/`cfexit`/`cfabort`. (The exception path already restored both depths via `TryHandler`.)

Regression coverage is `tests/tags/test_tags_cfexit_loop.cfm` — 14 assertions over body re-execution, single start phase, per-iteration `generatedContent`, persistent tag `variables`, caller writes reaching the next body pass, emission order, nested looping tags, the start-phase error, and the `cfbreak` case. It passes unchanged on both RustCFML and Lucee 7.0.4.

Closing as superseded by the shipped fix. Thank you for the work on this — the `cfbreak` analysis in particular pointed straight at the right design and away from the obvious wrong one.

PR #332, opened 2026-08-18
Add CFML tests: three gaps on the jwt-cfml surface the v0.606 crypto shims enabled

Blute:
## What

Three suites pinning what the v0.606.0 `Signature`/`KeyFactory` shims *unmasked*: real-world JWT verification now gets past the signature and dies (or silently corrupts) one layer further down. Found within the hour of pointing titan's Auth0 flow at v0.607.0.

1. **`tests/java_shims/test_java_date_shim_date_bifs.cfm`** — a `java.util.Date` shim instance must be usable *as a date* by the CFML date BIFs. The shim's own methods work (`getTime()` — pinned as a control), but `isDate()` is false and `dateAdd`/`dateDiff`/`dateCompare`/`dateTimeFormat` all throw `Invalid date: java.util.date`. jwt-cfml anchors its epoch claim conversion on exactly this object (`createObject('java','java.util.Date').init(javacast('int', 0))` + `dateAdd`/`dateDiff`), so every `decode()` of a token with an `exp` claim dies after the signature verifies. Assertions are timezone-agnostic (relative arithmetic; the one format leg accepts 1969/1970).

2. **`tests/stdlib/test_binarydecode_excess_padding.cfm`** — Lucee ignores surplus `=` in base64 entirely; RustCFML agrees for *partial*-group excess (`QQ===` → 1 byte, both engines — pinned as a control) but decodes a whole spurious `====` quad into a **trailing NUL byte** (`QUJD====` → 4 bytes vs Lucee's 3). jwt-cfml's `base64UrlToBinary` pads with `repeatString('=', 4 - (len % 4))` — i.e. four `=` when the length is already a multiple of 4 — so the lib's (admittedly buggy) idiom produces exactly this shape, and the NUL corrupts the decoded JWT header/payload JSON: `Invalid JSON: trailing characters`. Lucee's leniency is what the ecosystem has calibrated against.

3. **`tests/java_shims/test_java_util_base64.cfm`** — `java.util.Base64` is unshimmed. The conventional JWK→PEM finisher is `Base64.getEncoder().encodeToString(publicKey.getEncoded())`; a JWKS callback reaches this class right after `KeyFactory` succeeds. The suite is written resolution-first, so if/when the shim lands the encoder/decoder legs activate automatically (`encodeToString` ↔ `binaryEncode` parity, decoder round-trip).

## Why

These three are one story: v0.606 opened the road and the next three potholes became reachable. titan's login on v0.607.0 moved off its claims-only fallback automatically (`canVerifyAsymmetric()` went true) and then failed in the date conversion; fixing that exposed the NUL corruption; fixing *that* exposed the missing `Base64`. All three were patched app-side in pure CFML, so titan logs in with full JWKS/RS256 verification today — but every other Lucee codebase pointing jwt-cfml (or hand-rolled JWKS code) at RustCFML will hit the same sequence in the same order.

## Validation

- **Stock v0.607.0** (release binary, full runner): the three suites contribute exactly 9 red assertions — Date-shim 1/7 (the `getTime` control passes), padding 3/5 (both whole-quad legs red, the failure label prints the materialised NUL: `saw last byte: 0`), Base64 0/1. No other suite moves. (Amusing artifact: the NUL byte in the runner output makes grep treat the log as binary.)
- **Lucee 7.0** (docker `lucee/lucee:7.0`, repo as webroot, runner over HTTP): 7/7, 5/5, 3/3.

Test-only; no engine change suggested.

### DIFF
--- a/tests/java_shims/test_java_date_shim_date_bifs.cfm
+++ b/tests/java_shims/test_java_date_shim_date_bifs.cfm
@@ -1,0 +1,59 @@
+<cfscript>
+// A java.util.Date shim instance must be usable AS A DATE by the CFML date
+// BIFs — on Lucee (real JVM) isDate() is true and dateAdd/dateDiff/
+// dateCompare/dateTimeFormat all accept it. On RustCFML the shim exists and
+// its methods work (getTime() is fine), but every date BIF rejects it:
+//
+//   dateAdd:     Invalid date: java.util.date
+//   dateCompare: Invalid date1
+//
+// Repro class: this sits directly on the vendored jwt-cfml surface the
+// v0.606.0 Signature/KeyFactory shims enabled. jwt-cfml anchors epoch math
+// on `createObject('java','java.util.Date').init(javacast('int', 0))` and
+// converts exp/nbf claims via dateAdd/dateDiff against it — so the moment
+// canVerifyAsymmetric() went true, titan's Auth0 decode moved off its
+// fallback and died here instead: the shim unlocked a path the Date shim
+// then broke. (titan worked around it by rewriting the converters in pure
+// CFML; this pins the engine-side contract.)
+//
+// Assertions are timezone-agnostic (relative arithmetic, not absolute
+// formatting), measured on Lucee 7.0 (docker lucee/lucee:7.0).
+
+suiteBegin("java.util.Date shim instances work with the CFML date BIFs");
+
+epochBase = createObject("java", "java.util.Date").init(javacast("long", 0));
+
+// The shim itself is present and its methods work — that part is fine today.
+assert( "shim method getTime() works (control)", epochBase.getTime(), 0 );
+
+// isDate: a Date IS a date.
+assertTrue( "isDate() is true for a java.util.Date instance", isDate(epochBase) );
+
+// dateAdd accepts it and yields a real date.
+r1 = "(threw)";
+try { r1 = isDate(dateAdd("s", 60, epochBase)); } catch (any e) { r1 = "THREW: " & e.message; }
+assert( "dateAdd() accepts a java.util.Date base", r1, true );
+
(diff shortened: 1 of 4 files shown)


alexskinner:
Merged into v0.620.0 (222dc6fc) — thank you, this was an unusually good report, and "three potholes in the order you hit them" was exactly the right framing.

All three suites are in `main`, extended rather than replaced, credited to you in the test files and the commit. Every red is now green on both engines, because the gaps are fixed rather than pinned:

- **java.util.Date** stringified to its class name, which is why every date BIF rejected it. It now renders as the engine's own datetime form in local time (matching `Date.toString()`'s use of the default zone), so `isDate`/`dateAdd`/`dateDiff`/`dateCompare`/`dateTimeFormat` all accept it through the existing coercion path. One place, and your `getTime()` control still passes.
- **java.util.Base64** is shimmed — all six encoders/decoders, including the URL-safe alphabet JWTs actually use, `withoutPadding()`, and MIME's 76-character CRLF wrapping. Your resolution-first suite activated on its own; I added ten legs for the rest of the surface, all measured on Lucee.
- **binaryDecode padding**: `=` decodes as 0 and a group's first byte was always pushed, so a padding-only group fabricated the NUL. A group opening with padding now carries no bits, which also fixed `QUJD==`, `QUJD===`, `QQ====` and `QUJD====QUJD` — 19 padding shapes now byte-identical to Lucee. (The remaining rows in known-issues §54 are the separate malformed-input tolerance, unchanged.)

I could not merge the branch itself: registering the suites while they were still red would have put the release gate permanently in the red — that is the trip-wire that let a regression hide for 23 releases here once — so the tests had to land in the same commit as the fixes. The content is yours.

One thing your padding work turned up in passing: `arrayLen()` on a binary returns 0 here where Lucee returns the byte count, and `isArray` is false where Lucee says true. Filing separately.

PR #333, opened 2026-08-18
Add CFML tests: cfexecute quote parsing, cfqueryparam array list values (+ elvis error-scope pin)

Blute:
## What

Two gaps with the same shape — **Lucee coerces, RustCFML takes the value literally** — plus one green parity pin. All found running titan's full PDF pipeline and search screens on v0.607.0.

1. **`tests/tags/test_cfexecute_argument_quoting.cfm`** — Lucee tokenizes the `arguments` string shell-style: a double-quoted span is ONE argument with the quotes stripped. RustCFML splits on whitespace straight *through* the quotes and passes the quote characters to the program. The discriminating leg: `arguments='"two  spaces"'` must preserve its internal double space (one argv entry); RustCFML yields `"two spaces"` — split into `"two` + `spaces"`, rejoined by echo. (The existing `CfexecuteQuotedArgFixture` pins only that such source *parses* — it sits behind `<cfif false>`; this pins runtime semantics.)

2. **`tests/tags/test_cfqueryparam_list_array_value.cfm`** — `cfqueryparam list="true"` with an **array** value: Lucee expands one bind per element (`value="#someArray#"` inside `IN (...)` is a standard idiom). RustCFML binds the serialized array as a single parameter, and the failure mode differs by driver: **QoQ silently matches 0 rows** — no error, a search just looks empty — while PostgreSQL throws `cannot bind "[<uuid>…" as PostgreSQL uuid: invalid character: found `[` at 0`. Cross-engine QoQ legs carry the RED; the uuid leg is `RUSTCFML_TEST_PG_DS`-gated per convention.

3. **`tests/core/test_elvis_error_scope.cfm`** — a **green pin**: elvis (`?:`) swallows a *thrown* left side (not just null/undefined) identically on both engines now — thrown call, throw-inside-chain, undefined deep path, unknown function all yield the default. This was NOT true at v0.595.1, where a thrown left side propagated (the measurement footnote in #323); the convergence happened silently somewhere in v0.596–v0.607, so this pins it against re-divergence in either direction. Defensive one-liners (`cfg().setting ?: default`) lean on this scope constantly.

## Why

Repro classes, both from titan on v0.607.0 this week:

- **cfexecute**: titan renders every PDF by shelling out to the typst CLI with conventionally-quoted paths — and it turned out typst had *never* successfully run via cfexecute on RustCFML: it received `"/path/out.pdf"` quotes-included and died with `could not infer output format` (extension = `pdf"`). Every call site had to be rewritten unquoted with space-free paths — a workaround that collapses the moment a path legitimately contains a space.
- **cfqueryparam**: titan's pg_trgm search idiom binds `id IN (<cfqueryparam list="true" value="#idsInSearchTerm(...)#">)` where the helper returns an array of uuids — 21 call sites, every search screen. All broke (Postgres) or would silently empty (QoQ) until wrapped in `arrayToList()`.

## Validation

- **Stock v0.607.0** (release binary, full runner, PG container up): exactly 5 red assertions — cfexecute 1/3 (both quote legs, labels show the literal-quote output), array-list 1/4 (both QoQ legs at 0 rows + the live PG bind error), elvis pin 4/4 green. No other suite moves.
- **Lucee 7.0** (docker `lucee/lucee:7.0`, repo as webroot, runner over HTTP): 3/3, 4/4, 4/4.

## Related, filed separately

The same session also hit the S3 shim ignoring `Application.cfc`'s `this.s3` settings (its header comment notes the integration is "currently routed via env vars") — titan now aliases `S3_*` → `AWS_*`/`LUCEE_S3_HOST` in compose as a workaround. That one needs an application-context switch plus a live bucket to test cleanly, so I'm opening it as an issue rather than forcing a test shape.

Test-only; no engine change suggested.

### DIFF
--- a/tests/core/test_elvis_error_scope.cfm
+++ b/tests/core/test_elvis_error_scope.cfm
@@ -1,0 +1,39 @@
+<cfscript>
+// Elvis (?:) error scope — a PARITY PIN, currently green on both engines.
+//
+// On Lucee the left side of `?:` is evaluated in a swallow-everything scope:
+// not just null/undefined, but a genuinely THROWN exception anywhere in the
+// chain yields the default. Measured identical on RustCFML v0.607.0 — but it
+// has not always been: at v0.595.1 a thrown left side PROPAGATED here
+// (observed during the getBaseTagData work, PR #323's measurement footnote:
+// `getBaseTagData(...).attributes.marker ?: "x"` threw on RustCFML while
+// Lucee returned "x"). The behaviour converged silently somewhere in
+// v0.596–v0.607; this suite pins it so any future change is deliberate.
+//
+// Real-world weight: defensive one-liners lean on this scope constantly —
+// `cfg().setting ?: default` is written assuming a failing cfg() means
+// "use the default", not "500".
+
+suiteBegin("elvis (?:) swallows thrown left sides, not just null/undefined (Lucee parity)");
+
+function boomFn() { throw(message="boom"); }
+
+r1 = "(propagated)";
+try { r1 = boomFn() ?: "dflt"; } catch (any e) { r1 = "THREW: " & e.message; }
+assert( "a thrown function call yields the default", r1, "dflt" );
+
+r2 = "(propagated)";
+try { r2 = boomFn().member ?: "dflt"; } catch (any e) { r2 = "THREW: " & e.message; }
+assert( "a throw inside a call-then-member chain yields the default", r2, "dflt" );
+
+emptySt = {};
+r3 = "(propagated)";
+try { r3 = emptySt.a.b ?: "dflt"; } catch (any e) { r3 = "THREW: " & e.message; }
+assert( "an undefined deep path yields the default (the documented case)", r3, "dflt" );
+
+r4 = "(propagated)";
+try { r4 = totally_unknown_fn_xyz() ?: "dflt"; } catch (any e) { r4 = "THREW: " & e.message; }
+assert( "an unknown function name yields the default", r4, "dflt" );
+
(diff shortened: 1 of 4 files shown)


alexskinner:
Merged into v0.620.0 (901a96c7 and 5b62d3fa) — thank you. All three suites are in `main`, credited to you in the test files and the commits.

- **cfexecute** is fixed as GH #337: a faithful port of Lucee's `Command.toList`. Your `'"two  spaces"'` leg was the right discriminator. I added ten argv-boundary legs using `printf` rather than `echo`, which shows the exact argument boundaries `echo` hides by rejoining with single spaces — 15 tokenization cases now byte-identical to Lucee 7.1.0.204.
- **cfqueryparam list=true with an array** now expands one bind per element. One helper serves all five sites — the QoQ unwrapper and the four driver-side ones, including the placeholder counter, which has to agree with the expansion or the `?` count and the bind count drift apart. You were right that the silent QoQ leg is the dangerous one.

  Your QoQ legs alone would not have caught the driver bug, though: those are separate code paths, and the PostgreSQL failure you hit lives in the driver one. I added a SQLite datasource leg — it self-skips where SQLite will not load, so it does not report a missing Lucee extension as a divergence.
- **The elvis pin** is registered as-is and green on both engines.

I could not merge the branch itself: registering red suites would have put the release gate permanently in the red, so the tests had to land in the same commit as the fixes.

One thing worth knowing for future cross-engine tests: script-form `cfexecute()` is unsupported here (tag form works), and Lucee cannot call a BIF via `argumentCollection` with mixed-case argument names. Filing the first.

PR #362, opened 2026-08-25
Add CFML test: cfqueryparam collection crosses between threads sharing one component instance

Blute:
## What

`tests/database/test_cfqueryparam_shared_instance_threads.cfm` + fixture `tests/database/SharedQueryInstanceFixture.cfc` (and a `/database` mapping in `tests/Application.cfc` so the fixture resolves on Lucee, same reason as `/comments`).

One fixture instance, 8 `cfthread`s, 40 calls each. Every leg is caught, so a throwing engine reports counts rather than aborting the file. Each call returns the bound values (`i|tag|tag`) so a row is checked against what *this* caller bound, not just for the absence of an error.

- **Sequential control** — the helper shape, one caller, 40 calls: 0 errors, 0 wrong rows.
- **THE GAP — shared instance, method call inside the query body**: the `cfquery` body does `<cfset local.p = paramFor(...)/>` (a method on the *same* instance) between two `cfqueryparam`s. Expected 0 errors / 0 crossed rows. **Stock: 254 bind-count errors (`Wrong number of parameters passed to query. Got 4, needed 3`) and 44 rows that came back carrying another thread's bound values with no error raised.** The silent second mode is the one that matters — under load it persists one caller's values under another's parameters.
- **Control — same instance, same three placeholders, inline `cfqueryparam`s only**: clean on stock. So it is not "a shared instance running queries concurrently".
- **Control — helper shape, one instance per thread**: clean on stock. So it is not "a method call inside a query body". The accumulator is keyed to the component instance / call frame rather than to the executing request or thread.

## Why

Every framework's data layer is an application-scoped singleton, and building a `cfqueryparam` attribute struct through a mapper method *while the statement is being assembled* is a natural shape (moopa's `record_writer.updateRecord` → `writeMapper.buildQueryParam`, inside the `cfquery`). In titan a quote's line editor autosaves per keystroke; two overlapping requests through `db.save` produced `queryExecute: 6 positional parameter(s) supplied but the SQL only consumes 4` and `statement needs 20 placeholder(s) but only 3 of 3 supplied remain` — the sale UPDATE (4 slots) receiving the sale_line UPDATE's parameters and vice versa. The error points at the parameter count, nowhere near the cause, so it presents as a payload bug (it was first reported as "deleting text in the description breaks saving"). Sequentially the same payloads save fine; 12 parallel saves fail 12/12. Any two users saving anything at the same moment through a shared write path can hit it, and per the wrong-row leg, not always loudly.

## Shape

- Inline in-memory SQLite datasource (`{ class: "org.sqlite.JDBC", connectionString: "jdbc:sqlite::memory:" }`), the `test_duplicate_result_columns` / `test_dml_returns_empty_query` convention — no server, no PG. Probe-and-skip guard: where the engine has no SQLite driver the live legs are replaced by one informational pass.
- Concurrency is in-process via `cfthread` (join with timeout), so the runner exercises it on the CLI — no `--serve` needed. Thread names carry a per-run UUID so a rerun in the same request cannot collide.
- The fixture takes the datasource in `init()` so the identical component runs on both engines.
- All expected values are measured, not assumed: the failure counts above are from stock v0.629.0 (macOS aarch64 release binary) and vary run to run (262/32, 266/43, 254/44 across three runs); the assertions expect 0 so the numbers only appear in the failure message.

## Validation

```
FAIL | cfqueryparam collection is per-call under concurrent use of one component instance | 6/8 passed (2 failed)
       FAIL: shared instance + method call inside query body: 8 threads x 40 — no bind-count errors [first: queryExecute: query error: Wrong number of parameters passed to query. Got 4, needed 3] | expected: [0] | got: [254]
       FAIL: shared instance + method call inside query body: no row carries another thread's bound values | expected: [0] | got: [44]
SUMMARY: 8216/8218 passed across 838 suites
```

Baseline on `upstream/main` (a0030d2) with the same binary: `SUMMARY: 8210/8210 passed across 837 suites` — no other suite moved.

**Lucee 7.0.5.41** (`docker lucee/lucee:7.0`, repo as webroot, `sqlite-jdbc-3.46.1.3.jar` dropped in `lucee-server/context/lib` so every leg runs live rather than skipping):

```
PASS | cfqueryparam collection is per-call under concurrent use of one component instance | 8/8 passed
```

### DIFF
--- a/tests/Application.cfc
+++ b/tests/Application.cfc
@@ -30,6 +30,10 @@ component {
     // resolves (issue #69 fixture).
     this.mappings["/comments"] = getDirectoryFromPath(getCurrentTemplatePath()) & "comments/";
 
+    // Map "database" so createObject("component", "database.SharedQueryInstanceFixture")
+    // resolves on Lucee too (same reason as /comments above).
+    this.mappings["/database"] = getDirectoryFromPath(getCurrentTemplatePath()) & "database/";
+
     // Distinct mapping name (NOT a real webroot subdirectory) used by
     // tags/test_mapping_include.cfm to isolate this.mappings-based cfinclude
     // resolution: a `/wheelsmapprobe/...` include can ONLY be found via this
--- a/tests/database/SharedQueryInstanceFixture.cfc
+++ b/tests/database/SharedQueryInstanceFixture.cfc
@@ -1,0 +1,57 @@
+<!---
+    Fixture for test_cfqueryparam_shared_instance_threads.cfm.
+
+    One instance of this component is shared by several threads at once (the
+    application-scoped singleton shape: a framework's db/record-writer object
+    held in application.lib.db). Each method runs a tag-mode cfquery whose
+    placeholder set is fixed, then returns the bound values so the caller can
+    check the ROW as well as the absence of an error.
+--->
+<cfcomponent output="false">
+
+    <!--- The datasource is injected (an inline sqlite JDBC struct in the test)
+          so the same fixture runs on Lucee, which has no named `testds`. --->
+    <cffunction name="init" access="public" returntype="any" output="false">
+        <cfargument name="ds" type="any" required="true" />
+        <cfset variables.ds = arguments.ds />
+        <cfreturn this />
+    </cffunction>
+
+    <!--- Builds a cfqueryparam attribute struct. Called from INSIDE the query
+          body below, mid-way through accumulating the statement's parameters
+          (the moopa write_mapper.buildQueryParam shape). --->
+    <cffunction name="paramFor" access="public" returntype="struct" output="false">
+        <cfargument name="v" type="string" required="true" />
+        <cfreturn { cfsqltype: "varchar", value: arguments.v, null: false } />
+    </cffunction>
+
+    <!--- LEG A shape: the query body calls a method on this same instance while
+          its cfqueryparams are being collected. Three placeholders:
+          ?  ,  ?  ,  ?   ->  a | b1 | b2 --->
+    <cffunction name="selectViaHelper" access="public" returntype="string" output="false">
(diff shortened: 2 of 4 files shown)


alexskinner:
Fixed in v0.636.0 (016c0dc3) — suite in as
`tests/database/test_cfqueryparam_shared_instance_threads.cfm`, 8/8 here and on
Lucee, and stable across repeated runs (a race would show up).

This was the most serious report in the queue, for the reason you identified:
the loud mode is a confusing bind-count error, and the quiet mode persists one
caller's values under another caller's parameters.

Root cause. A `<cfquery>` whose body contains control flow can't have its SQL or
its param set known at compile time, so that shape lowers to a runtime
accumulator — an engine variable the body appends each `cfqueryparam` to, handed
to `queryExecute` at the end. The generated init is an **unscoped** assignment.
Inside a CFC method under classic localmode an unscoped write lands in the
component `variables` scope, which every thread sharing that instance also
holds. Two concurrent callers therefore appended into one array.

Your four legs isolate it exactly, and each control was load-bearing in
diagnosis: the inline control passes because no control flow means no runtime
accumulator at all, and the one-instance-per-thread control passes because each
instance has its own `variables`. I did briefly conclude from the first that the
trigger was merely "control flow in the body" and that the helper call was
incidental — building your no-helper variant disproved that in a minute. The
sibling method call is genuinely required: it is what pushes and pops a frame
whose `__variables` copy gets written back over the shared instance scope.

The accumulator is now pinned to the frame assembling the statement, so it is
per-invocation whatever the scope rules say. `__cfhttp_params` — the same
mechanism for runtime `<cfhttp>` bodies — had the identical exposure and is
covered by the same guard, though nobody had reported it.

PR #363, opened 2026-08-25
Add CFML test: <cf_name> body does not inherit the caller's output context (cfmodule does)

Blute:
## What

`tests/tags/test_customtag_prefix_body_output.cfm` + fixture `tests/tags/ctpathroot/ctpath_body_echo.cfm` (at the custom-tag-path root already declared in `tests/Application.cfc`, so no deep search involved). The fixture records `thisTag.generatedContent` in `request.ctpathBodyEcho` and emits nothing — the assertions read the request key, so no output capture is in play. Every leg is under `cftry`.

- **Control — `<cfmodule name="…">` inside the caller's `<cfoutput>`**: body evaluated, tag receives `VALUE`. Passes on stock.
- **Control — `<cfmodule template="…">`**: same. Passes on stock.
- **Control — `<cf_name>` with its *own* inner `<cfoutput>`**: passes on stock — the prefix path *can* evaluate a body.
- **GAP 1 — `<cf_name>#x#</cf_name>` inside the caller's `<cfoutput>`**: expected `VALUE`; **stock hands the tag the literal `#probeValue#`.**
- **GAP 2 — `<cf_name>#x#</cf_name>` inside a `<cffunction output="true">` body** (implicit output context, no cfoutput tag anywhere): expected `VALUE`; **stock: literal `#probeValue#`.**

So the gap is precisely "the prefix invocation path does not inherit the enclosing output context" — the same tag through either `cfmodule` form does.

## Why

Moopa/titan form-control tags wrap their Alpine component script in `<cf_once>` (a dedupe-once-per-request tag) inside the control's own `<cfoutput>`; the script builds image URLs from `#server.system.environment.TWIC_PICS_URL#`. On RustCFML every thumbnail `src` and "click to open" link became `https://host/page#server.system.environment.TWIC_PICS_URL/…` — broken images and a click that navigated to a fragment. The second flavour is every route whose `<cffunction output="true">` body is a `<cf_layout_default>…</cf_layout_default>` block: any `#expr#` directly in the layout body renders literally. 19 sites in that one app (a scan for `cf_` bodies relying on an enclosing output context); the app-side workaround is an explicit inner `<cfoutput>` at each, which is a no-op on Lucee.

## Shape

- Fixture at the custom-tag-path root: resolves on both engines without deep search (the deep-search suite covers that separately).
- Result via request scope rather than `cfsavecontent`: wrapping a custom tag *inside a function* in `cfsavecontent` drops the tag's output on **both** engines, which masked GAP 2 in a first draft — worth knowing if you reach for that shape.
- Expected values are measured on Lucee 7.0.5.41, not assumed.

## Validation

```
FAIL | Tags: a prefix-invoked custom tag body inherits the caller's output context | 3/5 passed (2 failed)
       FAIL: <cf_name> body inherits the caller's enclosing cfoutput | expected: [VALUE] | got: [#probeValue#]
       FAIL: <cf_name> body inherits a cffunction output=true context | expected: [VALUE] | got: [#probeValue#]
SUMMARY: 8213/8215 passed across 838 suites
```

```
PASS | Tags: a prefix-invoked custom tag body inherits the caller's output context | 5/5 passed
```

### DIFF
--- a/tests/runner.cfm
+++ b/tests/runner.cfm
@@ -509,6 +509,7 @@ include "harness.cfm";
 <!--- Fixture dirs under tests/tags/ctpathroot/; the path itself is declared --->
 <!--- in tests/Application.cfc. Reduced from the titan (Moopa) codebase port. --->
 <cf_runtest file="tags/test_customtag_path_deep_search.cfm">
+<cf_runtest file="tags/test_customtag_prefix_body_output.cfm">
 <cf_runtest file="tags/test_tags_buffer_recovery.cfm">
 <cf_runtest file="tags/test_tags_cfexecute.cfm">
 <cf_runtest file="tags/test_cfexecute_argument_quoting.cfm">
--- a/tests/tags/ctpathroot/ctpath_body_echo.cfm
+++ b/tests/tags/ctpathroot/ctpath_body_echo.cfm
@@ -1,0 +1,11 @@
+<!--- Fixture for tests/tags/test_customtag_prefix_body_output.cfm: record
+      exactly what the tag received as generatedContent in
+      request.ctpathBodyEcho and emit nothing (the caller asserts on the
+      request key, so no output capture is involved — cfsavecontent around
+      a custom tag inside a function drops the tag's output on BOTH engines,
+      which would mask the result). Lives at the custom-tag-path ROOT so it
+      resolves on both engines without deep search. --->
+<cfif thisTag.executionMode eq "end">
+    <cfset request.ctpathBodyEcho = trim(thisTag.generatedContent) />
+    <cfset thisTag.generatedContent = "" />
+</cfif>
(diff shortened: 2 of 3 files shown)


alexskinner:
Fixed in v0.636.0 (016c0dc3) — 5/5 here and on Lucee 7.1.0.204.

Your framing ("the prefix invocation path does not inherit the enclosing output
context") turned out to be the fix, almost literally. The tag preprocessor
threads an `in_cfoutput` flag through as it recurses into a body. The `cfmodule`
form and the `taglib` prefix form both pass the caller's flag down; the `cf_`
prefix form called the entry point that *restarts* at `in_cfoutput = false`. So
`#expr#` in the body was left as literal text while the same tag reached through
`<cfmodule>` interpolated it. One argument.

Your three controls are what made that diagnosable rather than a guess — control
3 in particular, since "the prefix path can evaluate a body, it just doesn't
inherit" is precisely the distinction between this bug and a much larger one.
GAP 2 needed no extra work once the flag was threaded, which confirms
`<cffunction output="true">` and an enclosing `<cfoutput>` are the same
mechanism internally.

Thanks also for the note about `cfsavecontent` dropping a custom tag's output
inside a function on both engines, and for routing the result through request
scope instead. I'd have reached for savecontent first and lost time to it.

This is the quiet kind: no error, just wrong markup — 19 sites in one app, and
every one of them a link or an image `src` that silently became a fragment.

PR #364, opened 2026-08-25
Add CFML test: hyphenated custom-tag attribute names (x-ref, data-foo) are parsed as expressions

Blute:
## What

`tests/tags/test_customtag_hyphenated_attributes.cfm` + fixture `tests/tags/ctpathroot/ctpath_attr_echo.cfm` (custom-tag-path root already declared in `tests/Application.cfc`; no deep search involved). The fixture records the sorted attribute names and the values of `x-ref` / `data-foo` in request scope and emits nothing. Every leg is under `cftry`, so the throw is asserted as a value.

- **Control — plain names** (`id`, `title`): tag runs, keys `id,title`. Passes on stock.
- **GAP 1 — `<cf_name id="m" x-ref="r1" />`**: expected keys `id,x-ref` with `attributes["x-ref"] = "r1"`. **Stock: throws `Variable 'x' is undefined` before the tag runs.**
- **GAP 2 — same via `<cfmodule name="…" x-ref="r2" />`**: **stock: same throw** — so it is the attribute parser, not the prefix-resolution path.
- **GAP 3 — `data-foo="d"`**: **stock: `Variable 'data' is undefined`** — any hyphenated name, not an Alpine/`x-*` special case.

## Why

Tags that wrap markup routinely take HTML-flavoured attribute names and read them with `attributes["x-data"]` / `<cfparam name="attributes['x-data']">` (the bracket-key `cfparam` spelling was pinned earlier in #318–#322; this is the caller side of the same idiom). In titan: `<cf_modal id="modal_cost_line" title="Cost Line" x-ref="modal_cost_line">` — the whole route 500s on RustCFML, and the reported stack line points at an unrelated part of the file, so it reads as a phantom variable bug.

## Shape

- Result via request scope rather than output capture (same reasoning as #363).
- Expected values measured on Lucee 7.0.5.41; keys are `listSort`ed so the assertion is order-independent across engines.

## Validation

```
FAIL | Tags: hyphenated attribute names on a custom tag are attribute keys, not expressions | 1/4 passed (3 failed)
       FAIL: <cf_name x-ref=...>: hyphenated name is an attribute key, value intact | expected: [id,x-ref | x-ref=r1 | data-foo=(absent)] | got: [THREW: Variable 'x' is undefined | x-ref= | data-foo=]
       FAIL: <cfmodule name= x-ref=...>: hyphenated name is an attribute key, value intact | expected: [id,x-ref | x-ref=r2 | data-foo=(absent)] | got: [THREW: Variable 'x' is undefined | x-ref= | data-foo=]
       FAIL: <cf_name data-foo=...>: any hyphenated name, not just x-* | expected: [data-foo,id | x-ref=(absent) | data-foo=d] | got: [THREW: Variable 'data' is undefined | x-ref= | data-foo=]
SUMMARY: 8211/8214 passed across 838 suites
```

```
PASS | Tags: hyphenated attribute names on a custom tag are attribute keys, not expressions | 4/4 passed
```

### DIFF
--- a/tests/runner.cfm
+++ b/tests/runner.cfm
@@ -509,6 +509,7 @@ include "harness.cfm";
 <!--- Fixture dirs under tests/tags/ctpathroot/; the path itself is declared --->
 <!--- in tests/Application.cfc. Reduced from the titan (Moopa) codebase port. --->
 <cf_runtest file="tags/test_customtag_path_deep_search.cfm">
+<cf_runtest file="tags/test_customtag_hyphenated_attributes.cfm">
 <cf_runtest file="tags/test_tags_buffer_recovery.cfm">
 <cf_runtest file="tags/test_tags_cfexecute.cfm">
 <cf_runtest file="tags/test_cfexecute_argument_quoting.cfm">
--- a/tests/tags/ctpathroot/ctpath_attr_echo.cfm
+++ b/tests/tags/ctpathroot/ctpath_attr_echo.cfm
@@ -1,0 +1,9 @@
+<!--- Fixture for tests/tags/test_customtag_hyphenated_attributes.cfm: record
+      the attribute names (sorted) and the value of the hyphenated ones in
+      request scope. Emits nothing. Lives at the custom-tag-path ROOT so it
+      resolves on both engines without deep search. --->
+<cfif thisTag.executionMode eq "start">
+    <cfset request.ctpathAttrKeys = listSort(structKeyList(attributes), "textnocase") />
+    <cfset request.ctpathAttrXref = structKeyExists(attributes, "x-ref") ? attributes["x-ref"] : "(absent)" />
+    <cfset request.ctpathAttrData = structKeyExists(attributes, "data-foo") ? attributes["data-foo"] : "(absent)" />
+</cfif>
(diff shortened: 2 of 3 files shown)


alexskinner:
Fixed in v0.636.0 (016c0dc3) — 4/4 here and on Lucee 7.1.0.204.

The attribute *parser* was never the problem: it reads a name up to `=` or
whitespace, so `x-ref` arrived intact. The bug was one step later, in how the
name was written back out. The custom-tag lowering builds a CFML struct literal
from the attributes and emitted the key bare, so `x-ref="r1"` became

```cfml
{ x-ref: "r1" }
```

which parses as the expression `x - ref` — hence `Variable 'x' is undefined`,
and hence the stack line pointing nowhere near the attribute, exactly as you
described. Any name that isn't a plain CFML identifier is now quoted as a
literal key.

Your GAP 2 was the useful half of the report: proving it reproduces through
`<cfmodule>` as well ruled out the prefix-resolution path and pointed straight
at the shared attribute-struct emission. There are three call sites that build
that literal — the `cf_` prefix form, `<cfmodule>`, and the `taglib` prefix form
— and all three are fixed, so the third one (which you didn't test) can't drift
away from the other two.

Built-in tags still emit keys bare. Their attribute vocabularies are fixed and
hyphen-free, and Lucee rejects unknown attributes on them anyway, so there's no
equivalent exposure there.

PR #365, opened 2026-08-25
Add CFML test: Lucee's serializeJSON date form ("August, 25 2026 09:00:14 +0000") is rejected by the date BIFs

Blute:
## What

`tests/stdlib/test_lucee_json_date_form.cfm`. Seven legs, every date call under try/catch so a throw is asserted as a value.

- **Controls** — `isDate("2026-08-25 09:00:14")` and `dateDiff` between two ISO-style strings: pass on stock. That form is what RustCFML's own `serializeJSON` emits, and both engines parse it.
- **GAP** — the string Lucee's `serializeJSON` emits for a date, `"August, 25 2026 09:00:14 +0000"` (measured on Lucee 7.0.5; also the offset-less variant older Lucee wrote):
  - `isDate(...)` → expected `true`; **stock `false`** (both variants)
  - `dateDiff("s", "2026-08-25 09:00:00", luceeForm)` → expected `14`; **stock throws `Invalid date2`**
  - `dateDiff("s", luceeForm, "2026-08-25 09:01:14")` → expected `60`; **stock throws `Invalid date1`**
  - `parseDateTime(luceeForm)` then `dateFormat`/`timeFormat` → expected `2026-08-25 09:00:14`; **stock throws `Cannot parse date: August, 25 2026 09:00:14`**

## Why

This is a data-compatibility gap rather than a code one, and it is one-directional. Anything written under Lucee into a JSON column with a raw CFML date inside carries Lucee's form forever; after switching engines every read of it throws. In titan the `zoho_oauth.tokens` jsonb row held `expires_when = dateAdd('s', expires_in, now())` from the Lucee era; on RustCFML the token-expiry check `dateDiff('s', now(), tokens.expires_when)` threw before the refresh that would have rewritten the row could run, so every Zoho integration call died with `{"success":false,"message":"Invalid date2"}`. A scan of all 27 jsonb columns in that app found two rows in that form — small blast radius there, but the shape ("serialised a struct with a date in it under Lucee") is common.

I've kept the serialiser's *output* format out of scope — RustCFML writing ISO is arguably the better choice and parses on both engines — this PR only pins that Lucee's form must be *readable*.

## Shape

- Pure stdlib, no fixtures, no datasource. Expected values measured on Lucee 7.0.5.41 (`serializeJSON({d: createDateTime(2026,8,25,9,0,14)})` → `{"D":"August, 25 2026 09:00:14 +0000"}`; `dateDiff` of the two strings above → `14`).
- Registered in `tests/runner.cfm` next to `stdlib/test_date_functions_extra.cfm`.

## Validation

```
FAIL | Dates: Lucee's serializeJSON date form round-trips through the date BIFs | 2/7 passed (5 failed)
       FAIL: isDate accepts Lucee's serializeJSON date form (with +0000) | expected truthy | got: [false]
       FAIL: isDate accepts Lucee's serializeJSON date form (no offset) | expected truthy | got: [false]
       FAIL: dateDiff with the Lucee form as date2 | expected: [14] | got: [THREW: Invalid date2]
       FAIL: dateDiff with the Lucee form as date1 | expected: [60] | got: [THREW: Invalid date1]
       FAIL: parseDateTime + dateFormat/timeFormat of the Lucee form names the same instant | expected: [2026-08-25 09:00:14] | got: [THREW: Cannot parse date: August, 25 2026 09:00:14]
SUMMARY: 8212/8217 passed across 838 suites
```

```
PASS | Dates: Lucee's serializeJSON date form round-trips through the date BIFs | 7/7 passed
```

### DIFF
--- a/tests/runner.cfm
+++ b/tests/runner.cfm
@@ -299,6 +299,7 @@ include "harness.cfm";
 <cf_runtest file="database/test_query_error_sqlstate_members.cfm">
 <cf_runtest file="database/test_pg_comment_placeholder_scan.cfm">
 <cf_runtest file="stdlib/test_date_functions_extra.cfm">
+<cf_runtest file="stdlib/test_lucee_json_date_form.cfm">
 <cf_runtest file="stdlib/test_locale_functions.cfm">
 <cf_runtest file="stdlib/test_java_i18n_shims.cfm">
 <cf_runtest file="stdlib/test_cache_functions.cfm">
--- a/tests/stdlib/test_lucee_json_date_form.cfm
+++ b/tests/stdlib/test_lucee_json_date_form.cfm
@@ -1,0 +1,50 @@
+<cfscript>
+suiteBegin("Dates: Lucee's serializeJSON date form round-trips through the date BIFs");
+
+// Lucee serialises a CFML date as "Month, dd yyyy HH:mm:ss +0000" — e.g.
+//   serializeJSON({d: createDateTime(2026,8,25,9,0,14)})
+//   -> {"D":"August, 25 2026 09:00:14 +0000"}           (Lucee 7.0.5, measured)
+// and its date parser accepts that string back: isDate() is true, and
+// dateDiff / parseDateTime / dateFormat all treat it as the date it names.
+//
+// RustCFML rejects the string: isDate() is false and every date BIF given it
+// throws "Invalid date2" (or the equivalent for its argument position). Its
+// own serializeJSON emits "2026-08-25 09:00:14", which BOTH engines parse, so
+// data written by RustCFML is fine everywhere — the gap is one-directional:
+// any JSON/jsonb column written under Lucee that holds a raw date is
+// unreadable after switching engines.
+//
+// Repro class: titan's zoho_oauth.tokens jsonb row, written under Lucee with
+// expires_when = dateAdd('s', expires_in, now()). On RustCFML the token
+// expiry check ran dateDiff('s', now(), tokens.expires_when) and every Zoho
+// integration call died with {"success":false,"message":"Invalid date2"} —
+// before the refresh that would have rewritten the row could run.
+//
+// Every leg is under try/catch; a throw is asserted as a value.
+
+luceeForm    = "August, 25 2026 09:00:14 +0000";  // measured serializeJSON output
+luceeFormNoZ = "August, 25 2026 09:00:14";         // same, sans offset (older Lucee)
+isoForm      = "2026-08-25 09:00:14";              // what RustCFML's serializeJSON writes; both parse
+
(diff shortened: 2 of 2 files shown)


alexskinner:
Fixed in v0.636.0 (016c0dc3), and this one earned its keep twice over — the
report was correct, and probing it against Lucee turned up a substantially worse
bug sitting behind it. **I did modify your test; please read the second half.**

## The reported gap

`"August, 25 2026 09:00:14 +0000"` puts the comma after the MONTH. Every
month-name pattern we had is `"%B %d, %Y"` — comma after the *day* — so nothing
matched and every date BIF rejected the string. Added, with and without the
trailing offset, for both full and abbreviated month names.

## The bug behind it

Working out what `+0000` should *mean*, I probed four offset-bearing forms
against Lucee 7.1.0.204 under `Europe/London` (UTC+1 in August). Before:

| input | Lucee | RustCFML (was) |
|---|---|---|
| `August, 25 2026 09:00:14 +0000` | `10:00:14` | *rejected* |
| `2026-08-25T09:00:14+00:00` | `10:00:14` | `09:00:14` |
| `2026-08-25T09:00:14Z` | `10:00:14` | `09:00:14` |
| `2026-08-25T09:00:14-05:00` | `15:00:14` | `09:00:14` |

**We discarded the offset on every offset-bearing form.** Lucee honours it and
expresses the result in the server timezone. A stored UTC timestamp read back on
a non-UTC server was wrong by exactly that server's offset, and six hours wrong
for `-05:00`. Zero difference wherever the server runs in UTC — CI, most
containers — and silently wrong everywhere else. Offsets are now honoured; all
four rows match Lucee exactly.

That is directly relevant to your repro, incidentally: the Zoho expiry check
would have been an hour off on a UK-hosted box even once the string parsed.

## Why your test changed

Your expected values were measured in a UTC container, where local == UTC, so
the offset being honoured or ignored is unobservable. Run against Lucee under
`Europe/London`, the suite as submitted fails **on Lucee itself**, 5/7 — the two
`dateDiff` legs off by exactly 3600s. It pinned a UTC-only reading of `+0000` as
though it were universal, and would have blocked the correct behaviour.

Rewritten timezone-independently: the offset-bearing legs now compare two
strings carrying the *same* offset (14s apart in any timezone), the offset-less
legs assert against the ISO form as before, and one new leg pins that the offset
is genuinely applied by comparing against `dateConvert("utc2Local", …)`, which
both engines agree on. 9/9 on RustCFML and on Lucee.

Worth carrying into the next one: an expected value measured only in a UTC
container can't distinguish "the offset was honoured" from "the offset was
thrown away". Your instinct to keep the serialiser's *output* format out of
scope was right, though — that half needed no change.

Issue #360, opened 2026-08-24
A CFC's methods stay visible as bare names in unrelated templates for the rest of the request


alexskinner:
Measured on **Lucee 7.1.0.204** and **RustCFML v0.627.0**. Instantiating a component publishes its method names into a table that later, unrelated templates resolve bare names against.

## Repro

`Widget.cfc`
```cfml
component {
	public string function someUniqueMethodName() { return "ran"; }
}
```

`a.cfm`
```cfml
<cfset w = new Widget()>
```

`b.cfm` — a different template, no reference to the component at all
```cfml
<cfscript>
try { p = someUniqueMethodName; writeOutput( "VISIBLE isFn=" & isCustomFunction( p ) ); }
catch ( any e ) { writeOutput( "not visible: " & e.message ); }
</cfscript>
```

`main.cfm`
```cfml
<cfinclude template="a.cfm"><cfinclude template="b.cfm">
```

| | result |
|---|---|
| Lucee | `not visible: variable [SOMEUNIQUEMETHODNAME] doesn't exist` |
| RustCFML | `VISIBLE isFn=true` |

The method is not on `variables` in `b.cfm` (`structKeyExists( variables, "someUniqueMethodName" )` is false) and `isDefined( "someUniqueMethodName" )` is false — only a bare-name *read* finds it, so the read path and the existence path disagree about the same name.

## Why it matters

A CFC's private naming choices become ambient names for every template that runs after it in the same request. A page that writes `status = ...` — or reads a bare `status` expecting its own variable — can silently pick up a component method of the same name instead, and the failure surfaces as a wrong value rather than an error. It also makes bare-name resolution order-dependent: the same template behaves differently depending on which components ran before it.

## How it surfaced

Fixing GH #351 (page scope has no `local` scope) removed the scope-view shortcut that used to mask this for the *specific* name `local`: `tests/oop/PresideFixLocalMethod.cfc` declares `function local()`, and after that component is constructed a later template's bare `local` resolved to that function. The leak is general, though — `normal()` on the same component leaks identically, and so does the `someUniqueMethodName` above. It predates that change.

`tests/core/test_local_at_page_scope.cfm` carries a comment pointing here, because it is why one assertion in that file is written against the returned *shape* rather than against the throw.

alexskinner:
Fixed in **v0.630.0** (f28d4b6).

[...]

Issue #234, opened 2026-07-02
cfthread: shared component instance mutations invisible after join; thread-context not detected inside thread body


alexskinner:
## Summary

Two related gaps around `<cfthread>` and shared component state, found via TestBox's own `StreamingService` concurrency specs. Related to the already-closed thread-scope issues #217/#219, but this is about mutating a shared **component instance** passed into a thread, not the `thread` scope itself.

## 1. Method calls made on a shared component instance from inside `cfthread` don't affect the instance as seen after `thread action="join"`

```cfml
var service = new testbox.system.util.StreamingService();

for ( var i = 1; i <= 10; i++ ) {
    thread name="queueThread_#i#" action="run" service="#service#" threadNum="#i#" {
        for ( var j = 1; j <= 5; j++ ) {
            attributes.service.queueEvent( "specEvent", { "thread": attributes.threadNum, "event": j } );
        }
    }
}
thread action="join" name="queueThread_1,queueThread_2,...,queueThread_10";

service.getQueuedEvents(); // Lucee/ACF: 50 events — rustcfml: 0 events
```

```
tests/specs/streaming/StreamingServiceTest.cfc
  "queue is thread-safe for concurrent writes"
  The expected length [50] is different than the actual length [0]
```

50 calls to `queueEvent()` across 10 threads produce **zero** queued events once observed from the parent thread after join — suggesting the `service` instance handed into each `cfthread` attribute is not the same shared reference the parent holds (e.g. cloned-on-attribute-pass rather than reference-shared), so mutations inside each thread are invisible outside it.

## 2. Code running inside a `cfthread` body can't detect that it's running inside a thread

```cfml
var service = new testbox.system.util.StreamingService();
service.setFlushEnabled( false );
service.setQueueMode( false ); // queueMode OFF, but streamEvent should still auto-queue when in a thread

thread name="autoQueueTest" action="run" service="#service#" {
    attributes.service.streamEvent( "specStart", { "id": "spec-thread" } );
}
thread action="join" name="autoQueueTest";

service.hasQueuedEvents(); // Lucee/ACF: true — rustcfml: false
```

```
tests/specs/streaming/StreamingServiceTest.cfc
  "streamEvent auto-queues events when called from a thread context"
  Expected [false] to be true
```

`StreamingService.streamEvent()` presumably calls something like `getCurrentThread()`/checks a thread-context flag to decide whether to force-queue rather than write straight to output; that detection appears to return false when actually running inside a `cfthread`.

[...]

PR #366, opened 2026-08-26
Test: a cfthread body inside a component method sees the component's methods

Blute:
## What
Pins Lucee's rule that a `<cfthread>` body written inside a CFC method runs in that component's context. Five legs against `tests/core/CfthreadMethodScopeFixture.cfc`, each returning `STATUS|out` (or `STATUS|THREW:message`) so stock's throw is asserted as a value, never an aborted file:

- **bare sibling private call** inside the thread body → `COMPLETED|10`
- **`variables.method`** inside the thread body → `COMPLETED|10`
- **bare call whose target itself calls further private siblings** (bare + `variables.`) → `COMPLETED|12`
- control: `this` passed as an attribute, public method called on it → `COMPLETED|10` (green on both)
- control: UDF passed as an attribute, aliased to a local then called bare → `COMPLETED|12` (green on both). Note Lucee's thread `attributes` is array-like, so `attributes.fn()` as a *member call* is not portable — the alias form is the one both engines accept.

## Why
v0.630.0 (GH #360) stopped publishing a component's method names into the ambient function table — right for *later templates*, but the thread body belongs to the component and lost them too. v0.629.0 resolved all three red legs. Real-world shape (titan): a route method kicks off a long SKU rebuild in a `cfthread` and calls the private worker from the body; on v0.634.0 every job now lands in the error state with `Variable 'rebuildSKUsNow' is undefined`.

## Shape
`tests/core/test_cfthread_component_method_scope.cfm` + fixture CFC beside it, registered in `tests/runner.cfm` after `core/test_function_scope_capture.cfm`. All expected values measured on Lucee 7.0 (`lucee/lucee:7.0`, repo as webroot).

## Validation
Stock v0.634.0 (release binary):
```
FAIL | cfthread inside a component method sees the component's methods | 2/5 passed (3 failed)
   FAIL: bare sibling private call resolves inside the thread body | expected: [COMPLETED|10] | got: [TERMINATED|THREW:Expression Error: Variable 'helper' is undefined]
   FAIL: variables.method resolves inside the thread body | expected: [COMPLETED|10] | got: [TERMINATED|THREW:Expression Error: Variable 'helper' is undefined]
   FAIL: bare call whose target itself calls further private siblings | expected: [COMPLETED|12] | got: [TERMINATED|THREW:Expression Error: Variable 'helperNested' is undefined]
FAIL | queryExecute failures are catchable as type=database | 3/4 passed (1 failed)
SUMMARY: 8534/8538 passed across 857 suites
FAILED:  4 assertion(s) in 2 suite(s)
```
Baseline before this suite: `SUMMARY: 8532/8533 passed across 856 suites` — the only other FAIL (queryExecute … type=database) is pre-existing; no other suite moved.

Lucee 7.0: `PASS | cfthread inside a component method sees the component's methods | 5/5 passed`.

### DIFF
--- a/tests/core/CfthreadMethodScopeFixture.cfc
+++ b/tests/core/CfthreadMethodScopeFixture.cfc
@@ -1,0 +1,79 @@
+<cfcomponent>
+    <!---
+        A cfthread body spawned from inside a component method runs with that
+        component's context: its methods (public AND private) resolve by bare
+        name and through `variables.` exactly as they do in the method itself.
+        Each leg returns "STATUS|out" (or "STATUS|THREW:message") so the caller
+        asserts a value rather than an aborted request.
+    --->
+
+    <cffunction name="bareSiblingCall" access="public">
+        <cfthread name="ctms_bare" x="5">
+            <cfset thread.out = helper(attributes.x) />
+        </cfthread>
+        <cfthread action="join" name="ctms_bare" timeout="5000" />
+        <cfreturn describe(cfthread.ctms_bare) />
+    </cffunction>
+
+    <cffunction name="variablesScopedCall" access="public">
+        <cfthread name="ctms_vars" x="5">
+            <cfset thread.out = variables.helper(attributes.x) />
+        </cfthread>
+        <cfthread action="join" name="ctms_vars" timeout="5000" />
+        <cfreturn describe(cfthread.ctms_vars) />
+    </cffunction>
+
+    <cffunction name="bareSiblingCallNested" access="public">
+        <cfthread name="ctms_nested" x="5">
+            <cfset thread.out = helperNested(attributes.x) />
+        </cfthread>
+        <cfthread action="join" name="ctms_nested" timeout="5000" />
+        <cfreturn describe(cfthread.ctms_nested) />
+    </cffunction>
+
+    <cffunction name="thisPassedAsAttribute" access="public">
+        <cfthread name="ctms_this" x="5" self="#this#">
+            <cfset thread.out = attributes.self.publicHelper(attributes.x) />
+        </cfthread>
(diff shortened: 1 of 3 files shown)


alexskinner:
Fixed in v0.636.0 (016c0dc3) — your five legs are in the suite as
`tests/core/test_cfthread_component_method_scope.cfm`, green here and on Lucee
7.1.0.204.

Your diagnosis pointed at the right release; the mechanism underneath it was one
layer down, and worth writing up because it had been latent for a long time.

`build_thread_seed` flattens the thread body's captured `__variables` scope into
per-thread by-value copies — that is the GH #234 fix, so sibling threads don't
all write into one shared live scope. It flattened with `iter()`. On a CFC
instance the method entries are **not in the instance's own map at all**: they
hang off it as a shared per-class flyweight table (the component-model memory
win — ~40 method entries live once per class instead of once per instance), and
plain iteration cannot see it. So the flatten faithfully carried the component's
DATA into the thread and silently dropped every one of its METHODS.

That was true long before v0.630.0. It was invisible because component methods
were *also* published into the ambient `user_functions` table, which the seed
copies wholesale — so the thread body found `helper` there instead. #360
correctly stopped publishing them, and took the thread body's last remaining
source of its own siblings with it.

Two changes: flatten via `snapshot_with_methods()` so the class table comes
along, and overlay the component scope onto the child VM's page scope, because
that is what `variables.` resolves against inside the body. The second half is
what your leg 2 needed — and also the case your legs did not reach, where the
body calls a sibling method that itself uses `variables.`. Both stay by-value
per thread, so #234 is untouched.

One note on your baseline: you reported 2/5 on stock v0.634.0, and that is what
I measured too. I briefly read 3/5 early on and it was a stale binary, not a
difference in our trees.

PR #369, opened 2026-08-28
Add CFML test: a component method named url() replaces the caller's url scope

Blute:
## What
Pins Lucee's rule that a component's **method name never touches the caller's scopes** — specifically a method named `url()`. Four shapes, each in its own served request against `tests/core/url_method_scope/probe.cfm` (so the broken scope cannot leak into the runner's request), asserted as `name=[value]` fields:

- **`createObject()` at page level of a CFC with `url()`** → `structKeyList(url)` still `shape,probe`, `url.probe` still `1`, and a subsequent **`url.written = "yes"` is visible** (stock: keys `[]`, `UNDEF`, `LOST`)
- **the same `createObject()` inside a function** → url scope intact inside the function *and* in the caller after it returns (stock: `[]` / `UNDEF`)
- control: an **argument** named `url` → scope undisturbed, binds normally (green on both)
- control: a method named **`form()`** → `form`/`url` scopes undisturbed (green on both) — the trigger is specifically a method named after the `url` scope

The method itself stays callable on both engines (`fn-url:k`); it's the caller's scope that goes missing for the rest of the request.

## Why
Real-world shape (titan/Moopa): a lib CFC is an S3 proxy exposing `url(key)`. Moopa instantiates every lib in `onApplicationStart`, so on the request that re-boots the application after a `?init=` `applicationStop()`, `url.route` (injected by the `.cfconfig.json` front-controller fallback) is gone by the time `onRequestStart` runs — `cgi.query_string` still says `route=/theme/forms`, `structIsEmpty(url)` is true and writes to `url` vanish. Moopa's `cfparam url.route default="/"` then dispatches home: every re-init landed on the dashboard. Bisected with a per-lib header dump; the scope flipped to empty right after the lib with `url()` loaded and stayed that way.

## Shape
`tests/core/test_component_method_named_url_scope.cfm` + `tests/core/url_method_scope/{probe.cfm, UrlMethodFixture.cfc, UrlArgFixture.cfc, FormMethodFixture.cfc}`, registered in `tests/runner.cfm` after `core/test_function_scope_capture.cfm`. HTTP round-trip via `cfhttp` to `cgi.server_port` (skips from the CLI, same as the session-namespace suite). All expected values measured on Lucee 7.0 (`lucee/lucee:7.0`, repo as webroot, served on 8888 so `cgi.server_port` loops back).

## Validation
Stock v0.636.0 (release image), served from repo root:
```
FAIL | A component method named url() does not replace the caller's url scope | 6/12 passed (6 failed)
   FAIL: url scope keys survive createObject() of a CFC with a url() method | expected: [shape,probe] | got: []
   FAIL: url.probe still readable after createObject() | expected: [1] | got: [UNDEF]
   FAIL: a url-scope write after createObject() is visible | expected: [yes] | got: [LOST]
   FAIL: url scope intact inside the function that created the CFC | expected: [shape,probe] | got: []
   FAIL: url scope intact in the caller after the function returns | expected: [shape,probe] | got: []
   FAIL: url.probe still readable after createObject() inside a function | expected: [1] | got: [UNDEF]
SUMMARY: 8667/8689 passed across 869 suites
```
Baseline (`upstream/main` without this suite, same image): `SUMMARY: 8661/8677 passed across 868 suites` — the 8 pre-existing ERRORED files and the `cfhttp name= / file=` FAIL are identical in both runs; no other suite moved.

Lucee 7.0: `PASS | A component method named url() does not replace the caller's url scope | 12/12 passed`.

### DIFF
--- a/tests/core/test_component_method_named_url_scope.cfm
+++ b/tests/core/test_component_method_named_url_scope.cfm
@@ -1,0 +1,78 @@
+<cfscript>
+suiteBegin("A component method named url() does not replace the caller's url scope");
+
+// ============================================================
+// Background
+// ============================================================
+// A CFC may declare a method called url() -- an S3/CDN proxy exposing
+// url(key) is the common shape. On Lucee the method name is just a member of
+// that component; the caller's url scope is untouched before, during and
+// after createObject()/the call, and a later `url.x = ...` write is visible.
+//
+// On RustCFML, merely instantiating such a component empties the caller's
+// url scope for the REST OF THE REQUEST: structKeyList(url) goes from
+// "probe,shape" to "", url.probe becomes undefined, and `url.written = "yes"`
+// silently vanishes. The same happens when the createObject() runs inside a
+// function. A method named form() does not disturb the form scope, and an
+// ARGUMENT named url is fine -- the trigger is specifically a method named
+// after the url scope.
+//
+// Real-world shape (titan/Moopa): the framework instantiates every lib CFC in
+// onApplicationStart; one is an S3 proxy with url(). On the request that
+// re-boots the application after applicationStop(), url.route (injected by
+// the front-controller fallback) is gone by the time onRequestStart runs, so
+// a "?init=" re-init redirects back to the page and then dispatches "/" --
+// every re-init lands on the home page.
+//
+// Each shape runs in its OWN request via cfhttp against probe.cfm, so the
+// broken scope cannot leak into the runner's request and poison later suites.
+// Runs only when served (cgi.server_port present); skips from the CLI.
+// ============================================================
+
+serverPort = structKeyExists(cgi, "server_port") ? trim(cgi.server_port) : "";
+skip = serverPort == "" || serverPort == "0";
+
+function urlMethodField(body, name) {
+    var m = reFind(name & "=\[([^\]]*)\]", body, 1, true);
+    if (arrayLen(m.pos) LT 2 || m.pos[2] EQ 0) return "MISSING";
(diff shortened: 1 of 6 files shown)


alexskinner:
Thanks @Blute — reproduced, root-caused and fixed in 06f48b06 on main.

**Cause.** A named function declaration compiles to `DefineFunction` + `StoreLocal(<name>)`. `StoreLocal`'s web-request-scope arm (`url`/`form`/`cgi`/`cookie`) committed *any* value to `self.globals[name]`, so declaring `function url()` replaced the live `url` scope struct with the method value itself. Merely instantiating the component emptied the caller's url scope for the rest of the request — reads went undefined and later `url.x = …` writes vanished. The `request`/`application`/`session` arms just above already gated on `if let CfmlValue::Struct`; the web-request arm did not. Fix is the same guard: only commit a struct.

**Two things your repro under-reported**, both now covered:

* The blast radius is all four web request scopes, not just `url`. Your `method_form` shape asserted only that the *url* scope survived — the `form` scope was in fact also being wiped by `FormMethodFixture`. I've added the missing `form_is_struct` assertion, plus a whole `method_cgi` shape (`cgi` is always populated, which makes it the strongest of the four).
* It reproduces from the CLI with no HTTP at all — the served request isn't needed to trigger it.

Your test file and fixtures went in essentially verbatim, extended from 12 to 17 assertions, and the header comment rewritten to past tense naming the real cause. Green on RustCFML (CLI, `--serve`, `--serve --production`) and on Lucee 7.1.

Closing in favour of the merged commit — the credit for finding it is yours.

Issue #302, opened 2026-08-02
v0.547.0 regression: server.system.environment reads empty on alternating workers — moopa boot fails "APP_NAME is required" (v0.546.0 clean; both modes, not just production)


Blute:
Same app image as #301 (only the engine version changed). **v0.546.0 is clean; v0.547.0 fails on every other request** — so this is 15c2c52 ("restore negative caching in the production component cache, now safe under the mappings-fingerprint key").

## Symptom

Alternating per-worker 500/200 on repeated requests, stable across container restarts:

```text
v0.547.0 + RUSTCFML_PRODUCTION=1 : 500 200 500 200 500 200
v0.547.0, no production mode     : 500 200 500 200 500 200   <-- dev mode affected too,
                                                                 despite the commit scoping
v0.546.0 (both modes)            : 302 200 200 200 200 ...
```

The 500s are moopa's boot guard:

```text
CFML error: Runtime Error: APP_NAME is required. Set APP_NAME to the app directory name under code/apps/ ...
```

## The failing expression is an env read, not a component probe

moopa `application.cfc:48`:

```cfml
<cfset application.app_name = trim(this.moopa_app_name ?: (server.system.environment.APP_NAME ?: "")) />
```

`APP_NAME=hub` is definitely set (docker `-e`; the alternating 200 workers read it fine and serve the app normally). On the failing workers `server.system.environment.APP_NAME` comes back empty during `onApplicationStart`, the guard throws, and that worker's application scope stays broken for its lifetime — hence the stable alternation.

So whatever 15c2c52 restored appears to be caching more than component paths: it's affecting `server.system.environment` (or the scope-chain/member resolution that backs it) per worker, and it's active without `RUSTCFML_PRODUCTION=1` as well.

Happy to retest a fix build against the same image as before — repro is deterministic. We're staying pinned on v0.546.0 meanwhile.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Blute:
Better characterization after more boots — **it's an idle-window race, not deterministic**, and my "alternating always" framing was incomplete:

| First request timing (v0.547.0) | Result (10 boots total, two different app configs) |
|---|---|
| Immediately when the listener opens (readiness-polled) | clean every time — warmup 302 then all 200s, zero errors |
| After the engine idles ~6–16s post-listen | alternating per-worker 500/200 every time, workers stay poisoned for their lifetime |

[...]

PR #383, opened 2026-08-30
Add CFML test: onApplicationStart gates all concurrent requests, not just the one that triggers it

Blute:
## What

CFML's application lifecycle contract: `onApplicationStart` runs to completion before **any** request executes against the new application scope. Lucee implements this as a gate — the request that triggers app start runs it, and every other request arriving mid-flight **blocks** until it finishes, then proceeds against the fully-built scope.

Stock gates only the triggering request. Concurrent requests see that the application scope exists and proceed straight into `onRequest` against whatever half-built state `onApplicationStart` has reached.

Five legs (the subtle ones bold):

1. Control: the triggering request runs `onApplicationStart`, blocks through it, sees `ready=true`.
2. **A request arriving at +400ms (mid-app-start) blocks and sees `ready=true` — never the half-built scope.**
3. **Same at +800ms.**
4. Control: steady-state request after app start sees `ready=true`.
5. **`onApplicationStart` ran exactly once — blocked requests must not re-trigger it.**

## Why

Real-world shape: an app whose `onApplicationStart` initialises singletons and a route registry progressively. On stock, every container (re)start under live traffic serves a window of failures staged by how far app start has progressed — first 500s ("Variable 'auth' is undefined": the request reached a `application.lib` that exists but doesn't hold `auth` yet), then 404s (route registry not yet loaded), then 200s. Every concurrent in-flight request during app start is another bypass, so ordinary healthcheck + proxy + user traffic widens the window: three 20ms pollers observed 344 × 500 across three restarts of a production app. The same three-poller probe against Lucee across a full boot: 2,688 samples, zero bypasses.

## Shape

`tests/lifecycle/test_application_start_serialisation.cfm`, registered in `tests/runner.cfm` beside the other application-lifecycle suites. Fixture app under `tests/lifecycle/application_start_gate/`: `onApplicationStart` sets `application.phase`, sleeps 1500ms, sets `application.ready`; `onRequest` reports `ready=true|MISSING` (or `runs=N` for leg 5). `this.name` is keyed on a per-run token passed by the test, so a warm server re-running the suite still exercises a cold app start. Probes are three `cfthread`s doing `cfhttp` to self at +0/+400/+800ms — same served-only skip guard (`cgi.server_port`) as the session-namespace suite. The 400/800ms offsets sit in the middle of the 1500ms sleep with wide margins on both sides.

## Validation

Stock v0.645.0 (release binary, repo as webroot):

```
FAIL | Requests serialise behind a running onApplicationStart | 3/5 passed (2 failed)
  - request at +400ms blocks behind app start and sees ready=true | expected: [ready=true] | got: [ready=MISSING]
  - request at +800ms blocks behind app start and sees ready=true | expected: [ready=true] | got: [ready=MISSING]
SUMMARY: 8842/8845 passed across 883 suites
```

Baseline (main without this test, same binary/host): `SUMMARY: 8839/8840 passed across 882 suites` — the only failing suite is the timezone-environment-dependent dates leg (`the +0000 offset is honoured | expected: [0] | got: [-3600]`, host TZ AEST), identical in both runs. No other suite moved.

Lucee 7.0.5.41 (`lucee/lucee:7.0`, repo as webroot, `curl tests/runner.cfm`):

```
PASS | Requests serialise behind a running onApplicationStart | 5/5 passed
```

### DIFF
--- a/tests/lifecycle/application_start_gate/Application.cfc
+++ b/tests/lifecycle/application_start_gate/Application.cfc
@@ -1,0 +1,23 @@
+component {
+	// Fresh application per test run: the runner passes a unique token so a
+	// warm server re-running the suite still exercises a cold app start.
+	this.name = "appstartgate_" & (url.app ?: "default");
+	this.sessionmanagement = false;
+
+	public boolean function onApplicationStart() {
+		application.runs = (application.runs ?: 0) + 1;
+		application.phase = "starting";
+		sleep(1500);
+		application.ready = true;
+		return true;
+	}
+
+	public boolean function onRequest(required string targetPage) output=true {
+		if ((url.op ?: "") == "runs") {
+			writeOutput("runs=" & (application.runs ?: 0));
+		} else {
+			writeOutput("ready=" & (application.ready ?: "MISSING"));
+		}
+		return true;
+	}
+}
--- a/tests/lifecycle/application_start_gate/page.cfm
+++ b/tests/lifecycle/application_start_gate/page.cfm
@@ -1,0 +1,1 @@
+<!--- Output is written by onRequest in Application.cfc --->
--- a/tests/lifecycle/test_application_start_serialisation.cfm
+++ b/tests/lifecycle/test_application_start_serialisation.cfm
@@ -1,0 +1,100 @@
+<cfscript>
+suiteBegin("Requests serialise behind a running onApplicationStart");
+
+// ============================================================
+// Background
+// ============================================================
+// CFML's application lifecycle contract: onApplicationStart runs to
+// completion before ANY request executes against the new application scope.
+// Lucee implements this as a gate -- the request that triggers app start runs
+// it, and every other request that arrives while it is mid-flight BLOCKS
+// until it finishes, then proceeds against the fully-built scope.
+//
+// RustCFML gates only the triggering request. Concurrent requests see that
+// the application scope exists and proceed straight into onRequest against
+// whatever half-built state onApplicationStart has reached. In a real app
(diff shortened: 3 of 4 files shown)


alexskinner:
The behaviour this test documents is fixed in [v0.646.0](https://github.com/RustCFML/RustCFML/releases/tag/v0.646.0) (43e01da) — the suite is green, 5/5.

Root cause was the ordering: `started = true` was flipped *before* `onApplicationStart` ran, not after. That did stop a second request re-firing the handler, but only by making it skip the whole start block — which is precisely the bypass this test describes. Every concurrent request saw a started application and went straight into `onRequest`.

Now gated on a per-application-name mutex, with `started` set only once the handler returns. Mutual exclusion plus a re-check under the gate gives the same once-only guarantee while actually making the second request wait, and it's released into a scope that's fully built rather than one that merely exists.

Details worth recording:

- The waiter needs no scope refresh — `application_scope` is a handle on the same `Arc` the starter mutates, so its writes are already visible. It does adopt the starter's function table, which its pre-start snapshot didn't have.
- `parking_lot::Mutex`, so a panic inside a user's `onApplicationStart` can't poison the gate for the life of the process.
- A start that re-enters application loading on its own thread skips the gate rather than deadlocking on a lock it already holds.
- A handler that calls `applicationStop()` is no longer marked started afterwards — that would have resurrected a stopped app as a started one with an empty scope.
- The GH #302 retry-on-failed-boot behaviour is preserved: the failure path now simply doesn't set the flag rather than rolling it back.

Issue #387, opened 2026-08-31
Linux: CFC lookup is case-sensitive (sqlRunner vs SqlRunner.cfc); Lucee is not


jjannek:
## Summary

On a case-sensitive filesystem (Linux ext4), rustcfml fails to resolve a CFC when the dotted component name's last segment does not match the on-disk filename case. Lucee resolves these. macOS APFS (default insensitive) hides the bug, which is why the Preside suite is green on jbookpro and HTTP 500 on Hetzner `agents`.

Architect lock: this is an **engine** bug. Do **not** rename Preside CFCs.

Hypothesis (non-binding): `exists_cached_path` / VFS `exists` do an exact-case path probe. Investigate and fix lookup; a request-scoped negative cache must not poison the case-folded hit.

Engine: v0.636.0 (`016c0dc`). Host: Linux, case-sensitive.

## Repro 1 — Preside conformance

Preside file on disk:

`system/services/database/SqlRunner.cfc`

Lookup:

`preside.system.services.database.sqlRunner`

On Linux rustcfml: `Could not find the component [preside.system.services.database.sqlRunner]` while executing `unit.api.forms.FormsServiceTest`. Suite aborts HTTP 500, no SUMMARY.

On jbookpro (case-insensitive): suite runs (signed 1825/17/0/2 on v0.636.0).

## Repro 2 — Preside app-smoke

On-disk:

`system/services/logger/TaskmanagerLogAppender.cfc`

Lookup:

`preside.system.services.logger.TaskManagerLogAppender`

Boot HTTP 500; `/`, `/admin`, `/admin/login` all stable 500s.

## Expected

Same as Lucee: component lookup is case-insensitive for the CFC filename (and dotted path segments) on case-sensitive filesystems. Exact-case hits should stay fast.

## Tiny isolated repro (should fail on Linux rustcfml, pass on Lucee)

```
mkdir -p /tmp/casecfc && cd /tmp/casecfc
cat > SqlRunner.cfc <<'CFC'
component { function ping() { return "ok"; } }
CFC
cat > index.cfm <<'CFM'
<cfset o = createObject("component", "sqlRunner")>
<cfoutput>#o.ping()#</cfoutput>
CFM
# rustcfml --serve this dir: createObject("component", "sqlRunner") should find SqlRunner.cfc
```

Second file: `TaskmanagerLogAppender.cfc` vs `createObject("component", "TaskManagerLogAppender")`.

jjannek:
[...]

PR #388, opened 2026-08-31
Fold every RealFs path segment (#387)

jjannek:
## Summary
- RealFs walks **every** path segment case-insensitively (not just the last, not only `.cfc`/`.cfm`).
- `storage/testDir/loading.gif` resolves to on-disk `storage/testdir/loading.gif`. Preside fixtures are **not** renamed.
- File BIFs (`fileRead`/`fileReadBinary`/`getFileInfo`/`fileExists`/`directoryExists`/`directoryList`) and `exists_cached_path` (via `vfs.exists`) use the same fold.

Fork PR: https://github.com/jjannek/RustCFML/pull/1
Commit: 8b29059


alexskinner (changes requested):
Thanks for digging into this — the Linux/case-sensitive breakage is real and worth fixing, and several parts of the implementation are careful (the fold is miss-only so exact-case stays fast; `resolve_file_bif_paths` runs *before* `sandbox_intercept`, so the sandbox sees the folded path and there's no bypass; the negative-cache poisoning the issue warned about is handled correctly; `tests/runner.cfm` is 8751/8751 green on the branch).

But I can't merge this as it stands. I built a case-sensitive APFS volume and ran the issue's own repros against Lucee 7.1.0.204 on the identical filesystem:

| | main | this PR | **Lucee 7.1.0.204** |
|---|---|---|---|
| `createObject("component","sqlRunner")` → `SqlRunner.cfc` | error | `ok` | **`invalid component definition, can't find component [sqlRunner]`** |
| same via a `/preside` mapping (Repro 1's exact shape) | error | `mapped-ok` | **same error** |
| `fileExists(".../Report.txt")`, on-disk `report.txt` | false | true | **false** |
| `fileRead(".../Report.txt")` | error | reads it | **`file or directory [...] does not exist`** |

So the premise doesn't hold. #387 says "Lucee resolves these" and this PR calls the helper `lucee_case_fold_path` / "Lucee-like", but on a case-sensitive filesystem stock Lucee 7 raises the very error the issue calls a RustCFML bug — for the relative form *and* the mapped form. That makes this a deliberate divergence *away* from Lucee in the permissive direction, not a compatibility fix. Before we settle on a behaviour, could you tell us your exact Lucee version and whether the Hetzner box has any non-default mapping/filesystem config? Something else is explaining the green Lucee run — their host, Preside's own resolution layer, or config — and it changes what the right fix is.

### Blocking issues

**1. It silently retargets destructive operations.** Measured with this branch's binary on the case-sensitive volume:

```
3 after fileWrite(Report.txt)     dir = report.txt,Report.txt
4 after fileDelete(REPORT.TXT)    dir = Report.txt          <- deleted report.txt
5 directoryDelete("./Cache",true) -> cache/ recursively gone
```

`filewrite`/`filewriteline`/`directorycreate` are correctly excluded via `allow_fallback=false`, but `filedelete`, `directorydelete`, `fileopen`, `filecopy`/`filemove` source and `filesetaccessmode` all fold. Lines 1 and 3 are the nasty pair: `fileExists` reports the file present while `fileWrite` creates a second one anyway, so `if (!fileExists(p)) fileWrite(p, ...)` now silently skips a legitimate write — and a delete can hit a file the caller never named.

**2. A shipped test fails on macOS.** `cargo test -p cfml-common --lib`:

```
test vfs::tests::lucee_case_fold_path_returns_on_disk_spelling ... FAILED
folded=.../storage/testDir/loading.gif
```

It asserts the result doesn't contain `testDir`, but on APFS the exact-path early return hands back the requested spelling. Per the verification gate in CLAUDE.md that's a release blocker on its own.

**3. The other tests are vacuous on a case-insensitive host.** I stubbed the fold out completely and all five `cfc_case_insensitive.rs` tests still passed. Combined with no `cargo test` in any workflow, nothing in this repo ever exercises the fix — it only runs on your Linux box.

**4. It reintroduces the uncached stat the existence cache exists to remove.** `resolve_existing_cf_path` calls `lucee_case_fold_path` *after* a cache hit, and that helper's first line is a raw `Path::exists()`. Every successful component/include resolution now pays a fresh stat, and on macOS/Windows it always returns the requested string unchanged, so it's pure waste there. I counted ~3,500-4,000 extra stats over one `tests/runner.cfm` run; warm-probe count per request went from 16.0 to 1.0 in an earlier campaign and this walks it back. Carry the folded spelling in the cache entry instead of re-deriving it.

### Smaller

- `lucee_case_fold_path` (raw `std::fs`) and `resolve_case_folded_cf_path` (VFS + memo) are two near-identical walkers that will drift.
- `RealFs::canonicalize` now folds. That's adjacent to the expandPath/symlink trap that bit us twice (v0.530.0, v0.630.1) — please leave `canonicalize` alone.

### What I think the shape should be

Narrow it to template/component resolution only: leave `RealFs` untouched so nothing leaks into file I/O semantics, drop the file-BIF and destructive-op folding entirely, do the fold in the VM behind the existence memo (with the folded spelling cached, and negative results cached too so the "override CFC absent" probes don't each pay a `read_dir`), and make the tests run against an injected case-sensitive VFS so they're meaningful on every host rather than only on Linux.

I'm going to push that version so you can see it concretely and tell me where it diverges from what you need on the Hetzner box. Thanks again for the repro and the isolated test case — those made this much quicker to pin down.

alexskinner:
Heads-up: the underlying issue (#387) is fixed and released in **v0.649.0**, so this PR is superseded. Leaving it open is your call, @jjannek — closing it costs nothing, but I did not want to close someone else's PR without saying why first.

What shipped keeps your core insight — **fold every path segment, not just the last** — and your `storage/testDir` case is covered. What changed:

* **Templates only.** File I/O stays case-sensitive, which is Lucee's actual behaviour on a case-sensitive filesystem. That matters because `filedelete`/`directorydelete`/`fileopen` share the same path-resolution list: measured on this branch, `fileDelete("./Foo.txt")` deleted an on-disk `foo.txt` and `directoryDelete("./Cache", true)` recursively removed `cache/`.
* **Two-pass rather than fold-in-place**, so priority survives — an exact match in the last mapping still beats a case-folded one in the first. The single-pass version silently loaded the wrong CFC on a tree where both spellings exist; there's a regression test for it now.
* **Retry on `compile_file_cached_req`'s error path**, so a template that reads pays nothing — no extra stat on the hot path.
* **Cached per directory**, not per path, behind the existence memo's two layers and its negative-answer generation, so many absent override probes in one directory cost one `read_dir`.
* Tests inject a case-sensitive VFS, so the fold is exercised on macOS and Windows too.

And thank you for the isolated repro in #387 — it's what made the Lucee comparison quick, which in turn is what found that the issue's premise needed correcting (stock Lucee 7 raises the same not-found; we ship a deliberate superset, recorded as known-issues §75).

alexskinner:
Closing this — but not before re-measuring it, because part of it turned out to still be a live bug.

I rebuilt the case-sensitive APFS volume and ran your shapes against **v0.685.2** and **Lucee 7.1.0.204** from the same docroot. One case you were right about had never actually been fixed:

| case (on-disk vs requested) | v0.685.2 | Lucee 7.1.0.204 |
|---|---|---|
| `views/Partial.cfm` ← `views/partial.cfm` (file mis-cased) | ok | throws |
| `views/Partial.cfm` ← **`Views/partial.cfm`** (dir mis-cased, relative) | **fails** | throws |
| `lvc/Partial.cfm` ← `/LVC/partial.cfm` (webroot-absolute) | ok | throws |
| `CaseOne.cfc` ← `caseone`, **CLI mode** | **fails** | throws |
| `CaseOne.cfc` ← `caseone`, serve mode | ok | throws |
| `stor/testdir/file.txt` ← `stor/testDir/file.txt` (`fileRead`) | throws | throws |

Both failures were one root cause. `fold_template_path` walks the path a segment at a time and bailed on `if parent.is_empty()` — so a *relative* path whose **first** segment is mis-cased had nothing accumulated yet and could never fold. An empty accumulator is the current directory, not a missing one. Every absolute shape already folded, which is why the suite stayed green: its fixtures are all absolute. v0.649.0's release note cited `cfinclude template="Views/partial.cfm"` → `views/Partial.cfm` as covered, and it was not — your PR would have caught that case and mine did not.

Fixed in **v0.685.3** (`c4e4d86b`), with a regression test that I checked **fails** without the engine change — the harness had the same empty-parent assumption, so the case was invisible from both sides.

**The rest of the PR stays rejected, and the measurement is what settles it rather than my earlier argument alone.** On the same volume Lucee folds *nothing* — every mis-cased spelling throws, including the file-I/O shapes. So `fileExists`/`fileRead`/`directoryList` staying case-sensitive is exact Lucee parity, and folding them (plus `RealFs`, plus `canonicalize`) is a divergence in the permissive direction that brings the destructive-op hazard with it: `fileDelete("./Foo.txt")` deleting an on-disk `foo.txt` is not a trade anyone can opt into, since it shares the resolution list. Template lookup has no such hazard — it only ever reads.

On the fixture that started that half: Preside's `FileSystemStorageProviderTest` hardcodes `/tests/resources/fileStorage/storage/testDir/loading.gif` while the committed fixture is `storage/testdir/`. Measured above, that `fileReadBinary` fails on **Lucee** on a case-sensitive filesystem too — so that test cannot pass on Lucee/Linux either. It's worth raising upstream against Preside (fix the path in the test, or rename the fixture); nothing the engine should paper over.

Thank you — genuinely. The core insight in this PR, *fold every segment and not just the last*, is what shipped, and your repro in #387 is what made the Lucee comparison quick enough to be worth trusting. The commit credits you by name. Recorded as known-issues §75.

PR #416, opened 2026-09-02
Bound expression recursion depth to prevent stack overflow on deeply nested source

chiliec:
## What

The recursive-descent expression parser has no depth limit. Deeply nested source — e.g. `<cfset x = ((((...))))>` or `#((((...))))#` — recurses through the full precedence chain (`parse_expression` → `parse_ternary` → … → `parse_primary` → back to `parse_expression` for the grouped sub-expression) once per nesting level, until the thread stack overflows and the process aborts:

```
fatal runtime error: stack overflow, aborting
```

On the default 8 MB main-thread stack this triggers at roughly **300 nested parens** — about 600 bytes of source. Since `Parser::parse` runs untrusted CFML (a template engine processes user-supplied templates/input), a tiny crafted source string is a denial of service.

## Reproduce (before this change)

```rust
let src = format!("<cfset x = {}1{}>", "(".repeat(400), ")".repeat(400));
let mut p = cfml_compiler::parser::Parser::new(src);
let _ = p.parse(); // stack overflow -> process aborts
```

## Fix

Track expression-recursion depth on the `Parser` and return a `ParseError` ("expression nesting too deep") once it exceeds `MAX_EXPR_DEPTH` (64) instead of recursing further. `parse_expression` is the single chokepoint every nesting level passes through, so the guard covers parens, operators, calls, etc.

64 is far above any hand-written CFML expression (each level walks ~10 precedence frames, so this stays comfortably within a normal stack). Normal expressions are unaffected — a realistic `(1 + 2) * (3 - 4) / (5 + 6)` and 30-deep nesting still parse, and all existing tests pass.

## Test

`tests/expr_depth_limit.rs` parses deeply nested input (10k / 100k deep) on the default 8 MB stack and asserts a returned error rather than an abort, plus a check that normal expressions still parse. On `main` this aborts; with the fix it passes. `cargo test -p cfml-compiler` (35 pass) and `cargo clippy` are clean.

---
*Disclosure: prepared with AI assistance; the crash was reproduced and the fix compiled/tested locally before submitting.*

### DIFF
--- a/crates/cfml-compiler/src/parser.rs
+++ b/crates/cfml-compiler/src/parser.rs
@@ -11,7 +11,18 @@ pub struct Parser {
     current: usize,
     /// Map from token index → javadoc comment text immediately preceding it.
     doc_comments: std::collections::HashMap<usize, String>,
+    /// Current expression-recursion depth, bounded by `MAX_EXPR_DEPTH` so a
+    /// deeply nested expression (e.g. `((((...))))`) returns a parse error
+    /// instead of overflowing the stack.
+    expr_depth: usize,
 }
+
+/// Maximum expression nesting depth. Deeply nested untrusted source would
+/// otherwise recurse until the stack overflows and the process aborts. Each
+/// nesting level walks the full precedence chain (~10 frames), so this stays
+/// well below what a thread stack can hold while remaining far above any
+/// hand-written CFML expression.
+const MAX_EXPR_DEPTH: usize = 64;
 
 #[derive(Debug)]
 pub struct ParseError {
@@ -29,6 +40,7 @@ impl Parser {
             tokens,
             current: 0,
             doc_comments,
+            expr_depth: 0,
         }
     }
 
@@ -5735,7 +5747,16 @@ impl Parser {
     // ---- Expression Parsing (Pratt-style precedence climbing) ----
 
     fn parse_expression(&mut self) -> Result<Expression, ParseError> {
-        self.parse_assignment_expr()
+        // Bound expression recursion so deeply nested input (e.g. `((((...))))`)
+        // returns a parse error instead of overflowing the stack.
+        self.expr_depth += 1;
+        if self.expr_depth > MAX_EXPR_DEPTH {
+            self.expr_depth -= 1;
+            return Err(self.parse_error("expression nesting too deep"));
+        }
+        let result = self.parse_assignment_expr();
+        self.expr_depth -= 1;
+        result
     }
 
     fn parse_assignment_expr(&mut self) -> Result<Expression, ParseError> {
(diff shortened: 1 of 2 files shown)


alexskinner:
Merged as a2cf40f — thank you, this was a well-scoped find and a clean fix.

I verified the report end to end before merging rather than taking the description at face value, and wanted to share what came out of that, since two details differ from the write-up.

**The crash reproduces exactly as described.** No question the vulnerability is real, and `parse_expression` is the right chokepoint for the expression half.

**The severity figure looks like a debug-build number.** The PR quotes "roughly 300 nested parens — about 600 bytes". On a release build with the real 8 MiB main-thread stack I measured the first abort at **~1600 parens (≈3.2 KB)**. Still a trivially cheap DoS, just not quite that cheap. Worth pinning down because it is the number that will get quoted later.

**`MAX_EXPR_DEPTH = 64` is well chosen, and now has evidence behind it.** The obvious risk with any fixed limit is rejecting legitimate code, so I instrumented the parser to record the maximum depth actually reached and ran it across **2,985 real files** (all of Preside plus our own `tests/`). The deepest expression anywhere is **9**. Nothing exceeds 16. Your limit has roughly 7x headroom over the worst real-world code, which is a comfortable place to be.

**One correction to the framing.** The description says `parse_expression` "is the single chokepoint every nesting level passes through, so the guard covers parens, operators, calls, etc." That is true for expressions, but not for the parser as a whole — several shapes still aborted with the patch applied:

| shape | first abort (release, 8 MiB) | recursion site |
|---|---|---|
| nested `<cfoutput>` | ~1,600 | `tags_to_script` — overflows *before* the parser runs |
| nested `function(){}` | ~1,600 | `parse_statement` |
| nested `<cfif>` / `if(){}` | ~3,200 | `parse_statement` |
| `!!!!…x`, `----…x` | ~5,000 (debug) | `parse_not` / `parse_unary` self-recursion |

The nested-tag one is the interesting case: it overflows inside the preprocessor, so no parser-side guard can reach it regardless of where the bound is placed.

I have followed up in 583f75b with two more bounds on the same pattern as yours — `MAX_STMT_DEPTH` in `parse_statement` (the single funnel for nested blocks and function declarations) and `MAX_TAG_NEST_DEPTH` in `tags_to_script_inner`, which records a preprocess error so an over-deep nest surfaces as a compile error rather than a silently truncated body. Both are 64, calibrated the same way: measured across those 2,985 files the deepest statement nesting is 11 and the deepest tag nesting is 4.

All six shapes now exit with a compile error instead of `fatal runtime error: stack overflow, aborting`. Gates green: `cargo test --workspace`, `tests/runner.cfm` 8886/8886 across 891 suites, and the wasm32 build.

Thanks again — appreciated the clear reproduction steps and the disclosure note, both of which made this quick to verify.

